PR #1, opened 2023-10-25
Add syft image workflow

mkosiarc:
This is about migrating our syft dockerfile from build-definitions to here. At the same time, the github workflow is set up for the building and pushing of the image. Automated updates will be handled by renovate bot.

### DIFF
--- a/.github/workflows/build-syft-image.yml
+++ b/.github/workflows/build-syft-image.yml
@@ -1,0 +1,51 @@
+name: Build syft image
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - syft.Dockerfile
+    
+  pull_request:
+    branches:
+      - main
+    paths:
+      - syft.Dockerfile
+
+
+env:
+  REGISTRY: quay.io/redhat-appstudio
+  IMAGE_NAME: syft
+  
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    - name: Extract version from Dockerfile
+      id: extract-version
+      run: echo "::set-output name=version::$(grep -oP '(?<=ARG syft_version=")[^"]*' syft.Dockerfile)"
+
+    - name: Build Image
+      id: build-image
+      uses: redhat-actions/buildah-build@v2
+      with:
+        image: ${{ env.IMAGE_NAME }}
+        tags: v${{ steps.extract-version.outputs.version }}
+        context: .
+        containerfiles: |
+            ./syft.Dockerfile
+
+    - name: Push to Quay
+      if: github.event_name == 'push'  # don't push image from PR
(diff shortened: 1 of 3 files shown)


mmorhun (approved):
LGTM

.github/workflows/build-syft-image.yml
```diff
@@ -0,0 +1,51 @@
+name: Build syft image
```
chmeliik, line 1: We want to build the image in RHTAP, not in GH workflows right? Do we have/can you file a story for that?

.github/workflows/build-syft-image.yml
```diff
@@ -0,0 +1,51 @@
...
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    - name: Extract version from Dockerfile
+      id: extract-version
+      run: echo "::set-output name=version::$(grep -oP '(?<=ARG syft_version=")[^"]*' syft.Dockerfile)"
```
chmeliik, line 31: Apparently this is deprecated https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

(see the [examples](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/#examples) for the new way to do things)

.github/workflows/build-syft-image.yml
```diff
@@ -0,0 +1,51 @@
...
+
+    - name: Push to Quay
+      if: github.event_name == 'push'  # don't push image from PR
+      uses: redhat-actions/push-to-registry@v2
+      with:
+        image: ${{ steps.build-image.outputs.image }}
+        tags: ${{ steps.build-image.outputs.tags }}
+        registry: ${{ env.REGISTRY }}
+        username: ${{ secrets.QUAY_USERNAME }}
+        password: ${{ secrets.QUAY_PASSWORD }}
```
chmeliik, line 51: Could you document where you got these in case we have to rotate them in the future?

.github/workflows/build-syft-image.yml
mkosiarc: We discussed this on monday with Mykola and Shoubik, and for now this is the easier option. Other teams currently use github workflows for building as well. Eventually we want to migrate to RHTAP itself though.

.github/workflows/build-syft-image.yml
mkosiarc: This is something for @mmorhun

.github/workflows/build-syft-image.yml
chmeliik: Huh, ok. I see we like our build pipeline :smile:

.github/workflows/build-syft-image.yml
chmeliik: Usual for RHTAP? Are they documented somewhere?

PR #20, opened 2024-01-17
Fix source image utils tagging

chmeliik:
Previously, we tagged the image only with the 'latest' tag.

In quay.io, an image that doesn't have any tags cannot be pulled. By
pushing to the 'latest' tag, we would immediately invalidate the
previous image, breaking everyone who relied on the image.

Tag the image with the git commit sha to make sure every image will
always have at least one tag associated with it.

### DIFF
--- a/.github/workflows/source-build.yaml
+++ b/.github/workflows/source-build.yaml
@@ -46,7 +46,9 @@ jobs:
       uses: redhat-actions/buildah-build@v2
       with:
         image: ${{ env.IMAGE_NAME }}
-        tags: latest
+        tags: |
+          latest
+          ${{ github.sha }}
         context: ./source-container-build
         containerfiles: |
           ./source-container-build/Dockerfile
--- a/source-container-build/Makefile
+++ b/source-container-build/Makefile
@@ -5,10 +5,6 @@ IMAGE = quay.io/$(ORG)/build-definitions-source-image-build-utils:latest
 image:
 	podman build -t $(IMAGE) .
 
-.PHONY: push
-push:
-	podman push $(IMAGE)
-
 .PHONY: clean
 clean:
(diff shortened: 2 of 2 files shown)


chmeliik:
the GH workflow doesn't want to run for some reason :confused:

.github/workflows/source-build.yaml
```diff
@@ -46,11 +46,17 @@ jobs:
...
-        tags: latest
+        tags: |
+          latest
+          ${{ github.sha }}
         context: ./source-container-build
         containerfiles: |
           ./source-container-build/Dockerfile
 
+    - name: test
+      script:
```
mmorhun, line 57: I guess, this is invalid, so the GH action haven't run.

chmeliik:
PTAL - the changes will be fully tested only after this is merged, but I'm fairly confident it will work

The Makefile change is there mostly to cause the workflow to run :smile:

mkosiarc (approved):

source-container-build/Makefile
```diff
@@ -5,10 +5,6 @@ IMAGE = quay.io/$(ORG)/build-definitions-source-image-build-utils:latest
 image:
 	podman build -t $(IMAGE) .
 
-.PHONY: push
```
mmorhun, line 8: I suppose you restore it before merge :)

mmorhun (approved):

source-container-build/Makefile
chmeliik: It's not actually used in the CI, and I don't think running `make push` locally is a good idea

So I was going to actually remove it

tkdchen: This target is not for pushing to the official registry, just for local testing. Without correct credential, pushing to redhat-appstudio does not work actually.

PR #23, opened 2024-01-18
Don't build for and update the latest tag

tkdchen:
The latest tag should not change at all. And subsequent builds for newer updates to source build should have their own tag. Then, any pipeline referencing the latest will not be broken. The new image reference can be updated by Renovate to the build-definitions.


tkdchen:
This is not required. So, close it.

PR #27, opened 2024-01-24
Tag source image based on cosign convention

tkdchen:
STONEBLD-1945

The old image scheme ${binary_image}.src is still pushed for
backward compatibility, which will be removed eventually.

Requires #26 to be merged in advance.

### DIFF
--- a/source-container-build/app/source_build.py
+++ b/source-container-build/app/source_build.py
@@ -352,7 +352,7 @@ def build_and_push(
     work_dir: str,
     sib_dirs: SourceImageBuildDirectories,
     bsi_script: str,
-    dest_image: str,
+    dest_images: list[str],
     build_result: BuildResult,
 ) -> None:
     log = logging.getLogger("source-build.build-and-push")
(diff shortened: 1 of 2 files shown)


source-container-build/app/source_build.py
```diff
@@ -381,23 +381,45 @@ def build_and_push(
...
     try:
         with open(digest_file, "r") as f:
             build_result["image_digest"] = f.read().strip()
     finally:
         os.unlink(digest_file)
 
 
+def fetch_image_manifest_digest(image: str) -> str:
+    cmd = ["skopeo", "inspect", "--format", "{{.Digest}}", f"docker://{image}"]
+    return run(cmd, check=True, text=True, capture_output=True).stdout.strip()
```
chmeliik, line 404: The default `skopeo inspect` can be very slow. `--no-tags` can help a bit:
```bash
$ time skopeo inspect --format '{{.Digest}}' docker://registry.access.redhat.com/ubi9/ubi-minimal
sha256:a340f4b9fb261a75c84666a3dccb88e193a116da3cebabaf9bcdc33609b61172
skopeo inspect --format '{{.Digest}}'   0.07s user 0.02s system 5% cpu 1.904 total

$ time skopeo inspect --format '{{.Digest}}' --no-tags docker://registry.access.redhat.com/ubi9/ubi-minimal
sha256:a340f4b9fb261a75c84666a3dccb88e193a116da3cebabaf9bcdc33609b61172
skopeo inspect --format '{{.Digest}}' --no-tags   0.07s user 0.02s system 7% cpu 1.256 total
```

And `--raw` is much faster:

```bash
$ time skopeo inspect --raw docker://registry.access.redhat.com/ubi9/ubi-minimal | sha256sum
a340f4b9fb261a75c84666a3dccb88e193a116da3cebabaf9bcdc33609b61172  -
skopeo inspect --raw docker://registry.access.redhat.com/ubi9/ubi-minimal  0.06s user 0.01s system 27% cpu 0.276 total
sha256sum  0.00s user 0.00s system 0% cpu 0.276 total
```

source-container-build/app/source_build.py
```diff
@@ -381,23 +381,45 @@ def build_and_push(
...
+
+    :param image: str, represent the built image.
+    :return: list of generated source container images.
+    """
+    # For backward-compatibility. It will be removed in near future.
+    deprecated_image = f"{image}.src"
+
+    # in format: sha256:1234567
+    digest = fetch_image_manifest_digest(image)
+    triangulate_image = f"{image.split(':')[0]}:{digest.replace(':', '-')}.src"
```
chmeliik, line 418: Splitting on `:` is technically wrong, since the registry can include a port number: `image-registry.openshift-image-registry.svc:5000`

chmeliik:
> Tag source image like cosign triangulate

`cosign triangulate` is not the command that tags images, it's the one that finds the `sha256-<digest>.sig` image

I think it would be more accurate to say that we're tagging the source image based on the cosign convention

source-container-build/app/source_build.py
tkdchen: Updated.

source-container-build/app/source_build.py
tkdchen: Update the command with `--no-tags` and still rely on skopeo to calculate the digest.

tkdchen:
> I think it would be more accurate to say that we're tagging the source image based on the cosign convention

Commit subject is updated.

chmeliik (approved):
LGTM :+1:

tkdchen:
PR 26 has been merged. This PR is rebased on that.

mmorhun (approved):

PR #33, opened 2024-02-04
Resolve source image

tkdchen:
STONEBLD-1874

A parent image pullspec within FROM instruction can be specified in
various ways. For example, tag could be latest, specifying digest only,
or giving both tag and digest. This patch inspects image's version and
release and checks source image with tag pattern `version-release-source`.
This applies to most of the ubi images and layered product images built
on top of ubi image.

Since source-build task consumes `BASE_IMAGES_DIGESTS` generated in
build-container task and each line of the image pullspecs has both tag and
digest, tag is removed and just keeping digest, which fixes two major
issues:

* skopeo-inspect cannot work with an image pullspec having both tag and
  digest.
* strings `<none>` and `latest` could be included.

By the way, the digest could be reused in a way instead of retrieving it
separately from registry.


source-container-build/app/source_build.py
```diff
@@ -417,6 +422,49 @@ def generate_source_images(image: str) -> list[str]:
...
+    """
+    parts = image.split("@")
+    digest = ""
+    if len(parts) == 2:
+        digest = parts[1]
+    parts = parts[0].rsplit(":", 1)
+    has_tag = len(parts) == 2 or (len(parts) > 2 and "/" in parts[0])
+    if digest and has_tag:
+        # Drop the tag
+        return parts[0] + "@" + digest
```
chmeliik, line 464: Does `parts[0] + "@" + digest` correctly handle the case where `len(parts) > 2`?

source-container-build/app/source_build.py
```diff
@@ -417,6 +422,49 @@ def generate_source_images(image: str) -> list[str]:
...
+    image_config = fetch_image_config(binary_image)
+    config_data = json.loads(image_config)
+    version = config_data["config"]["Labels"].get("version")
+    release = config_data["config"]["Labels"].get("release")
+    if not (version and release):
+        log.warning("Image %s is not labelled with version and release.", binary_image)
+        return
+    # Remove possible tag or digest from binary image
+    image_name = binary_image.split("@")[0]
+    image_name = image_name.rsplit(":", 1)[0]
```
chmeliik, line 443: Should we use `drop_tag` here as well?

source-container-build/app/test_source_build.py
```diff
@@ -448,15 +450,22 @@ def test_do_nothing_with_unsupported_registry(self):
...
+        skopeo_inspect_raw_rv.returncode = 1
+        skopeo_inspect_get_digest_rv = Mock()
+        skopeo_inspect_get_digest_rv.stdout = "sha256:123"
+        run.side_effect = [
+            # can't find out source image by version and release
+            skopeo_inspect_config_rv,
+            skopeo_inspect_raw_rv,
+            # can't find out source image by binary image digest
+            skopeo_inspect_get_digest_rv,
+            skopeo_inspect_raw_rv,
```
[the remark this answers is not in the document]
tkdchen: Removed.

source-container-build/app/source_build.py
tkdchen: Dropped `drop_tag` and replaced it with a rough image name parser so that the parsed result can be reused, at where your [comment](https://github.com/redhat-appstudio/build-tasks-dockerfiles/pull/33#discussion_r1482723897) mentioned.

source-container-build/app/test_source_build.py
```diff
@@ -1271,15 +1271,14 @@ def test_source_image_is_resolved(self, run: MagicMock):
...
-        ["ubuntu", "ubuntu"],
-        ["reg:3000", "reg:3000"],
-        ["reg:3000/img", "reg:3000/img"],
-        ["reg:3000/img:9.3", "reg:3000/img:9.3"],
-        ["reg:3000/img:9.3@sha256:123", "reg:3000/img@sha256:123"],
-        ["reg/org/img:9.3@sha256:123", "reg/org/img@sha256:123"],
-        ["reg/org/path/img:9.3", "reg/org/path/img:9.3"],
-        ["reg/org/path/img:9.3@sha256:123", "reg/org/path/img@sha256:123"],
+        ["ubuntu", ("ubuntu", "", "")],
+        ["reg:3000", ("reg", "3000", "")],
```
chmeliik, line 1270: Shouldn't the expected result here be `("reg:3000", "", "")`

tkdchen: This is an interesting case actually. Both parts `reg` and `3000` are valid for name and tag, even if `.` is included. They are usually recognized as that by tools as well, like skopeo and podman. And the first part is recognized as registry only when it includes `.` and `/` is included in the name. Anyway, this is just a rough parser and not to be a generic parser for image name.

chmeliik: ok

chmeliik (approved):
LGTM :+1:

PR #40, opened 2024-02-26
Add registry allowlist

tkdchen:
STONEBLD-1956

A new command line argument --registry-allowlist is added to make the registry configurable. Multiple registries can be specified via the argument, one per line.

The source-build task has to be updated to use the new CLI argument together with the Renovate update PR which updates image digest.

### DIFF
--- a/source-container-build/app/source_build.py
+++ b/source-container-build/app/source_build.py
@@ -11,11 +11,11 @@ import stat
 import sys
 import tarfile
 import tempfile
-import urllib.parse
 import filetype
 from dataclasses import dataclass, field
 from subprocess import run
 from typing import TypedDict, NotRequired, Literal, Final
+from urllib.parse import urlparse
 
 """
 Requires: git, skopeo, tar, BuildSourceImage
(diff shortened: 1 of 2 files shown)


source-container-build/app/source_build.py
```diff
@@ -491,8 +495,22 @@ def build(args) -> BuildResult:
...
         base_image = base_images[-1]
-        prepared = prepare_base_image_sources(base_image, work_dir, sib_dirs)
-        build_result["base_image_source_included"] = prepared
+
+        allowed = any(
+            [
+                urlparse("docker://" + base_image).netloc == registry
+                for registry in args.registry_allowlist
+            ]
+        )
```
chmeliik, line 504:
```suggestion
        allowed = urlparse("docker://" + base_image).netloc in args.registry_allowlist
```

chmeliik (approved):
LGTM, just a suggestion for simplification

mkosiarc (approved):

source-container-build/app/source_build.py
tkdchen: Good catch! Updated.

PR #46, opened 2024-03-06
Add script for generating base image sboms info

mkosiarc:
This script is then used in the buildah task step <TODO>

STONEBLD-2042


base-images-sbom-script/base_image_sbom_script.py
```diff
@@ -0,0 +1,86 @@
...
+
+from packageurl import PackageURL
+
+
+def get_base_images_sbom_data(images, is_last_from_scratch):
+
+    new_data = {"formulation": [{"components": []}]}
+
+    for index, image in enumerate(images):
+        property_name = f"builder-{index}"
```
chmeliik, line 13: You can use the same property name multiple times, there's no need to have the index in the name (that would make the property a bit harder to process)

Maybe `is_builder_image:for_stage`

base-images-sbom-script/base_image_sbom_script.py
```diff
@@ -0,0 +1,86 @@
...
+            }
+        )
+        
+        component = {
+            "type": "container",
+            "name": repository_without_tag,
+            "purl": purl.to_string(),
+            "properties": [{
+                "name": property_name,
+                "value": index
```
chmeliik, line 39: For base images, IMO the value can just be `"true"`

Also note that property values must be strings

base-images-sbom-script/base_image_sbom_script.py
```diff
@@ -0,0 +1,86 @@
...
+                "tag": tag,
+            }
+        )
+        
+        component = {
+            "type": "container",
+            "name": repository_without_tag,
+            "purl": purl.to_string(),
+            "properties": [{
+                "name": property_name,
```
chmeliik, line 38: We should probably use property names in the "Konflux namespace", e.g. `konflux:container:is_base_image`, `konflux:container:is_builder_image:for_stage`

base-images-sbom-script/base_image_sbom_script.py
```diff
@@ -0,0 +1,86 @@
...
+    is_last_from_scratch = False
+    if raw_images[-1] == "scratch":
+       is_last_from_scratch = True
+
+    
+    with args.sbom.open('r') as f:
+        sbom = json.load(f)
+        
+    new_data = get_base_images_sbom_data(digest_images, is_last_from_scratch)
+    sbom.update(new_data)
```
chmeliik, line 77: If the SBOM already had a `formulation` attribute, this would overwrite its content completely. I think we just want to append a `{"components": ...}` item to the formulations

base-images-sbom-script/base_image_sbom_script.py
mkosiarc: updated

base-images-sbom-script/base_image_sbom_script.py
mkosiarc: updated

base-images-sbom-script/base_image_sbom_script.py
mkosiarc: updated

base-images-sbom-script/base_image_sbom_script.py
mkosiarc: Yeah, it make sense to use a better approach here. But formulation is an array, so if it would exist beforehand, then there is also a possibility to have more items in it. That might create additional problem and also, the individual formulation item could potentially have a "components" already. So how much should we prepare here for these kinds of options?

base-images-sbom-script/app/tests/requirements-test.txt
```diff
@@ -0,0 +1,2 @@
+pytest
+pytest-mock
```
tkdchen, line 2: This lib looks pretty good.

base-images-sbom-script/app/requirements.txt
```diff
@@ -0,0 +1 @@
+packageurl-python
```
tkdchen, line 1: Pin all requirements?

base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -0,0 +1,112 @@
...
+        # example image: registry.access.redhat.com/ubi8/ubi:latest@sha256:627867e53ad6846afba2dfbf5cef1d54c868a9025633ef0afd546278d4654eac
+        # repository = registry.access.redhat.com/ubi8/ubi:latest
+        # digest = sha256:627867e53ad6846afba2dfbf5cef1d54c868a9025633ef0afd546278d4654eac
+        # repository_without_tag = registry.access.redhat.com/ubi8/ubi
+        # tag = latest
+        # name = ubi
+        repository, digest = image.split("@")
+        repository_without_tag, tag = repository.split(":")
+        # last fragment of the repository
+        name = repository_without_tag.split("/")[-1]
```
tkdchen, line 38: Tag can be `<none>`

The registry part can be in form `host:port`. 

Tag can be searched from the right side of the string, like https://github.com/tkdchen/my_image_utils/blob/main/src/my_image_utils/__init__.py#L139, or extract it with a regex that matches `:[0-9a-zA-Z._-]+$`.

Images built by Konflux have multiple path components in repository name, for example:
`quay.io/redhat-user-workloads/rh-acs-tenant/acs/collector:d92c93ab73f81442d99870528442855c75b535d7`

base-images-sbom-script/base_image_sbom_script.py
mkosiarc: Yeah you are right, there are actually no problems with this approach. Updated the code and added unit tests covering the scenarios where the input sbom did or did not have the formulation key

base-images-sbom-script/app/base_images_sbom_script.py
mkosiarc: Ok, so the fix was rather simple and this allowed to keep the whole functionality simple.
I extracted the image parsing to a function and added unit tests for it, so it's clear to see how it works and to verify it's functionality.

But if you still see some potential issues or would like to extend the testing, please let me know

base-images-sbom-script/README.md
```diff
@@ -0,0 +1,4 @@
+# base images sbom script
+
+This is a script that creates sbom data for base images. It takes list of base images and updates the provided sbom file
+with the data about them. This script is meant to be used in the buildah task in a separate step.
```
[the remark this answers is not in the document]
mkosiarc: updated the readme, should be a bit better. I will update them with instructions on how to run tests locally soon

base-images-sbom-script/README.md
```diff
@@ -0,0 +1,60 @@
...
+        {
+            "components": [
+                {
+                    "type": "container",
+                    "name": "quay.io/mkosiarc_rhtap/single-container-app",
+                    "purl": "pkg:oci/single-container-app@sha256:8f99627e843e931846855c5d899901bf093f5093e613a92745696a26b5420941?repository_url=quay.io/mkosiarc_rhtap/single-container-app&tag=f2566ab",
+                    "properties": [
+                        {
+                            "name": "builder-0",
+                            "value": 0
```
chmeliik, line 29: The property name needs an update to match the code

base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -0,0 +1,136 @@
...
+        sbom = json.load(f)
+
+    base_images_sbom_components = get_base_images_sbom_components(base_images_digests, is_last_from_scratch)
+    if "formulation" in sbom:
+        sbom["formulation"].append({"components": base_images_sbom_components})
+    else:
+        sbom.update({"formulation": [{"components": base_images_sbom_components}]})
+
+    with args.sbom.open("w") as f:
+        json.dump(sbom, f, indent=4)
```
chmeliik, line 132: Is there a specific reason to indent with 4 spaces? IMO we can use the default json.dump behavior to keep the SBOM file smaller.

(Or even use `json.dumps(separators=(',', ':'))` to make it even smaller)

base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -0,0 +1,136 @@
...
+            },
+        )
+
+        component = {
+            "type": "container",
+            "name": parsed_image.repository_without_tag,
+            "purl": purl.to_string(),
+            "properties": [{"name": property_name, "value": property_value}],
+        }
+        components.append(component)
```
chmeliik, line 78: Technically, it's possible to use the same image (same `repository_without_tag` and `digest`) in more than one build stage, or to use it both in a build and in the last stage.

In such a case, IMO we should report the image only once and combine its properties, e.g.

```json
{
  "type": "container",
  "name": "registry.access.redhat.com/ubi8/ubi",
  "purl": "pkg:oci/ubi@sha256:627867e53ad6846afba2dfbf5cef1d54c868a9025633ef0afd546278d4654eac?repository_url=registry.access.redhat.com/ubi8/ubi&tag=latest",
  "properties": [
    {
      "name": "konflux:container:is_builder_image:for_stage",
      "value": "0"
    },
    {
      "name": "konflux:container:is_base_image",
      "value": "true"
    }
  ]
}
```

Reporting the same component twice would not be correct because AFAIK purls must be unique.

base-images-sbom-script/app/requirements.txt
```diff
@@ -0,0 +1 @@
+packageurl-python==0.14.0
```
chmeliik, line 1: Could we use the `requirements.in` + `requirements.txt` approach (generate txt from in via `pip-compile`) to ensure all our dependencies are pinned?

base-images-sbom-script/app/base_images_sbom_script.py
chmeliik: And now that I think about it, maybe we don't need to include the `tag` in the purl - otherwise we'd have to think about what to do if the same image appears twice but with different tags

base-images-sbom-script/README.md
mkosiarc: updated

base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -0,0 +1,136 @@
...
+
+    with args.base_images_digests.open("r") as f:
+        base_images_digests = f.read().splitlines()
+
+    is_last_from_scratch = False
+    if base_images_from_dockerfile[-1] == "scratch":
+        is_last_from_scratch = True
+
+    with args.sbom.open("r") as f:
+        sbom = json.load(f)
```
tkdchen, line 123: minor suggestion for reducing the number of lines:

```python
base_images_from_dockerfile = args.base_images_from_dockerfile.read_text().splitlines()
base_images_digests = args.base_images_digests.read_text().splitlines()
sbom = json.load(args.sbom.read_bytes())
```

base-images-sbom-script/app/base_images_sbom_script.py
chmeliik: Ack, this seems fine then

base-images-sbom-script/app/base_images_sbom_script.py
mkosiarc: Happy to use the first two lines. But the sbom part did not work out of the box, and also we are also writing to the file later, so I kept it to avoid further changes.

base-images-sbom-script/app/base_images_sbom_script.py
mkosiarc: Removed the tags and added code for handling reuse of base images together with tests and I mentioned it in the README as well

base-images-sbom-script/app/requirements.txt
mkosiarc: updated

chmeliik (approved):
LGTM :+1:

tkdchen (approved):
:+1:

PR #48, opened 2024-03-07
Source rpm

skabashnyuk:
### What does this PR do?
Copy `*.src.rpm` files downloaded by cachi2 tool to `SourceImageBuildDirectories.rpm_dir` directory

### Screenshot/screencast of this PR
Знімок екрана 2024-03-07 о 10 22 11

Знімок екрана 2024-03-07 о 10 20 08

### What issues does this PR fix or reference?
 -  https://issues.redhat.com/browse/SVPI-757
 - https://issues.redhat.com/browse/KONFLUX-2076

### How to test this PR?
 - Reconfigure the prefetch task to use beta version of cachi2 with rpm.lock support https://github.com/containerbuildsystem/cachi2/compare/main...onosek:cachi2:rpm_prefetch  like so https://github.com/skabashnyuk/rpm-build/blob/4655cf23c0a30899e873335f3fca4831391d7cd7/.tekton/vim-push.yaml#L187
- enable `hermetic` build like https://github.com/skabashnyuk/rpm-build/blob/4655cf23c0a30899e873335f3fca4831391d7cd7/.tekton/vim-push.yaml#L29C1-L34C20
- enable `prefetch-input` as `rpm` like https://github.com/skabashnyuk/rpm-build/blob/4655cf23c0a30899e873335f3fca4831391d7cd7/.tekton/vim-push.yaml#L29C1-L34C20
- enable ` build-source-image` like https://github.com/skabashnyuk/rpm-build/blob/4655cf23c0a30899e873335f3fca4831391d7cd7/.tekton/vim-push.yaml#L29C1-L34C20
- Reconfigure task-source-build like so https://github.com/skabashnyuk/rpm-build/blob/4655cf23c0a30899e873335f3fca4831391d7cd7/.tekton/vim-push.yaml#L247
 
Resulted image : 
 - `quay.io/redhat-user-workloads/skabashn-tenant/rpm-build/vim@sha256:823adb17bb86e3151e0bd1e462d07b40d69181d091936e30e8d60c4dc7d940b4`
 - https://quay.io/repository/redhat-user-workloads/skabashn-tenant/rpm-build/vim?tab=tags&tag=latest

### DIFF
--- a/source-container-build/app/source_build.py
+++ b/source-container-build/app/source_build.py
@@ -12,6 +12,7 @@ import sys
 import tarfile
 import tempfile
 import filetype
+import hashlib
 from dataclasses import dataclass, field
 from subprocess import run
 from typing import TypedDict, NotRequired, Literal, Final
@@ -173,6 +174,15 @@ def fetch_image_manifest_digest(image: str) -> str:
     return run(cmd, check=True, text=True, capture_output=True).stdout.strip()
 
 
+# produces an artifact name that includes artifact's architecture
+# and repository id in the name
+def unique_srpm_artifact_name(file: str) -> str:
+    root, filename = os.path.split(file)
+    with open(file, "rb") as f:
+        digest = hashlib.file_digest(f, "sha256").hexdigest()
+        return f"{digest}-{filename}"
+
+
 def create_dir(*components) -> str:
     path = os.path.join(*components)
     os.makedirs(path)
@@ -316,6 +326,16 @@ def gather_prefetched_sources(
                 if mimetype and mimetype in ARCHIVE_MIMETYPES:
                     yield root, filename
 
+    def _find_prefetch_srpm_archives():
+        guess_mime = filetype.guess_mime
+        for root, dirs, files in os.walk(cachi2_output_dir):
+            dirs.sort()
+            for filename in sorted(files):
+                if filename.endswith(".src.rpm"):
+                    mimetype = guess_mime(os.path.join(root, filename))
+                    if mimetype and mimetype == "application/x-rpm":
+                        yield root, filename
+
     source_counter = itertools.count()
     prepared_sources_dir = create_dir(work_dir, "prefetched_sources")
     relative_to = os.path.relpath
@@ -331,7 +351,23 @@ def gather_prefetched_sources(
         shutil.copy(src, dest)
         sib_dirs.extra_src_dirs.append(f"{prepared_sources_dir}/{src_dir}")
 
-    gathered = next(source_counter) > 0
+    srpm_counter = itertools.count()
+    for root, filename in _find_prefetch_srpm_archives():
+        next(srpm_counter)
+        src = os.path.join(root, filename)
+        dest = os.path.join(sib_dirs.rpm_dir, filename)
+        if os.path.isfile(dest):
+            unique_src_name = unique_srpm_artifact_name(src)
+            unique_dest_name = unique_srpm_artifact_name(dest)
+            if unique_src_name != unique_dest_name:
+                dest = os.path.join(sib_dirs.rpm_dir, unique_src_name)
+            else:
+                log.debug("identical artifacts found for %s. skipping %s", dest, src)
+                continue
+        log.debug("copy prefetched rpm source %s to %s", src, dest)
+        shutil.copy(src, dest)
+
+    gathered = next(source_counter) + next(srpm_counter) > 0
     if not gathered:
         log.info("There is no prefetched source archive.")
 
--- a/source-container-build/app/test_source_build.py
+++ b/source-container-build/app/test_source_build.py
@@ -126,6 +126,19 @@ def create_fake_dep_packages(cachi2_output_dir: str, deps: list[str]) -> None:
                 f.write("any data")
 
 
+def create_fake_dep_packages_with_content(cachi2_output_dir: str, deps: dict[str, bytes]) -> None:
+    for package, content in deps.items():
+        pkg_mgr_path, filename = os.path.split(package)
+        packages_dir = os.path.join(cachi2_output_dir, "deps", pkg_mgr_path)
+        os.makedirs(packages_dir, exist_ok=True)
+        if filename.endswith(".rpm"):
+            with open(os.path.join(packages_dir, filename), "wb") as f:
+                f.write(b"\xed\xab\xee\xdb" + content)
+        else:
+            with open(os.path.join(packages_dir, filename), "wb") as f:
+                f.write(content)
+
+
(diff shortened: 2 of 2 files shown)


source-container-build/app/source_build.py
```diff
@@ -331,6 +340,15 @@ def _find_prefetch_source_archives():
...
         sib_dirs.extra_src_dirs.append(f"{prepared_sources_dir}/{src_dir}")
 
+    for root, filename in _find_prefetch_srpm_archives():
+        src_dir = f"src-{next(source_counter)}"
+        copy_dest_dir = f"{prepared_sources_dir}/{src_dir}/{relative_to(root, cachi2_output_dir)}"
+        os.makedirs(copy_dest_dir)
+        src = f"{root}/{filename}"
+        dest = sib_dirs.rpm_dir
+        log.debug("copy prefetched rpm source %s to %s", src, dest)
+        shutil.copy(src, dest)
```
chmeliik, line 350: IIRC, per the cachi2 design, all RPMs including SRPMs will be stored in architecture-specific directories, which means the same SRPM will likely exist in multiple paths.

Here, we de-duplicate implicitly (by filename). I'm wondering if we need a stronger check, e.g. check that SRPMs with the same filename are really byte-for-byte identical. If they're not, we essentially have three options:
* keep only one of them, document this as a known corner case
* fail the source build, consider this unsupported
* rename one of them and keep both

source-container-build/app/test_source_build.py
```diff
@@ -783,6 +796,29 @@ def test_gather_go_deps(self):
...
+            "output/x86_64/updates/repodata/6339e0d504567dc14512f01324d07bbb123-filelists.xml.zst",
+            "output/x86_64/updates-source/vim-9.1.113-1.fc38.src.rpm",
+            "output/x86_64/fedora-source/gpm.spec",
+            "output/x86_64/fedora-source/gpm-1.20.7-rhbz-668480-gpm-types-7-manpage-fixes.patch",
+            "output/x86_64/fedora-source/gpm.service",
+            "output/x86_64/fedora-source/gpm-1.20.7-42.fc38.src.rpm",
+            "output/x86_64/fedora-source/gpm-1.20.1-lib-silent.patch",
+            "output/x86_64/fedora-source/gpm-1.20.7.tar.xz",
+            "output/x86_64/fedora-source/1.20.7...e82d1a653ca94aa4ed12441424da6ce780b1e530.diff",
+            "output/x86_64/fedora-source/gpm-1.20.1-weak-wgetch.patch",
```
tkdchen, line 818: Does rpm prefetch implementation extract the content of SRPMs?

source-container-build/app/source_build.py
chmeliik: See this thread in the design: https://github.com/containerbuildsystem/cachi2/discussions/465#discussioncomment-8492211

In particular:

> (Source RPMs are not necessarily arch independent, you could have Sources or Patches conditional on architecture. Fedora prohibits that by policy.)

chmeliik: Let's do it :+1:

source-container-build/app/source_build.py
```diff
@@ -331,6 +340,13 @@ def _find_prefetch_source_archives():
         shutil.copy(src, dest)
         sib_dirs.extra_src_dirs.append(f"{prepared_sources_dir}/{src_dir}")
 
+    for root, filename in _find_prefetch_srpm_archives():
+        next(source_counter)
```
chmeliik, line 344: IIRC the source_counter is used for the names of non-SRPM sources (those that go into the `extra_src_dirs`. There's no need to increment the counter when we encounter an SRPM.

source-container-build/app/test_source_build.py
onosek: No, it does not. It just downloads them and stores them in the storage.
From rpm files, headers are read afterwards to get enough information for generating SBOM - and it is not even related to SRPMs, just RPMs.

onosek: I realized I made a mistake in the demonstration version of rpm prefetcher.
Directories containing sources should be together in "sources [1]":

```
output/<arch1>/<repoid>/Packages/*.rpm
output/<arch1>/sources/*.src.rpm

```
[1] https://github.com/containerbuildsystem/cachi2/discussions/465

source-container-build/app/source_build.py
```diff
@@ -173,6 +173,16 @@ def fetch_image_manifest_digest(image: str) -> str:
     return run(cmd, check=True, text=True, capture_output=True).stdout.strip()
 
 
+# produces an artifact name that includes artifact's architecture
+# and repository id in the name
+def unique_srp_artifact_name(file: str) -> str:
```
tkdchen, line 178: `s/srp/srpm/`

source-container-build/app/source_build.py
```diff
@@ -173,6 +173,16 @@ def fetch_image_manifest_digest(image: str) -> str:
...
 
 
+# produces an artifact name that includes artifact's architecture
+# and repository id in the name
+def unique_srpm_artifact_name(file: str) -> str:
+    root, filename = os.path.split(file)
+    repoid = os.path.basename(root)
+    head, tail = os.path.split(root)
+    arch = os.path.basename(head)
+    return f"{arch}-{repoid}-{filename}"
```
chmeliik, line 183: The source build script already adds SRPMs gathered from the source container of the base image. Those do not follow this naming scheme, they just have the normal filenames.

IMO we should stick to that as much as possible, only rename if necessary. (It's necessary to rename only if an SRPM with the same filename already exists and isn't the same file)

chmeliik, line 183: To avoid depending too much on the cachi2 output struture, IMO we can use something like `{checksum}-{filename}`. That should also better de-duplicate SRPMs that _should_ be de-duplicated

source-container-build/app/source_build.py
```diff
@@ -173,6 +174,15 @@ def fetch_image_manifest_digest(image: str) -> str:
     return run(cmd, check=True, text=True, capture_output=True).stdout.strip()
 
 
+# produces an artifact name that includes artifact's architecture
+# and repository id in the name
+def unique_srpm_artifact_name(file: str) -> str:
+    root, filename = os.path.split(file)
+    with open(file, "rb", buffering=0) as f:
```
chmeliik, line 181: For my understanding, why disable the the buffering here?

chmeliik (commented):
LGTM

Before merging, please squash the commits into one (or as many as you think appropriate) and add a more descriptive commit message (the first line of the PR description is a pretty good summary)

chmeliik (approved):

PR #54, opened 2024-03-20
feat: Support git submodules for source containers

ralphbean:
We found that the bootc project wished to use a git submodule to make managing their downstream repo's relationship to their upstream repo easy.

This change makes it so that the source archive of the git repository includes submodule content.

### DIFF
--- a/source-container-build/app/source_build.py
+++ b/source-container-build/app/source_build.py
@@ -391,9 +391,37 @@ def make_source_archive(
     repo_info = get_repo_info(source_dir)
     name_sha = f"{repo_info['name']}-{repo_info['last_commit_sha']}"
     output_archive = f"{source_archive_dir}/{name_sha}.tar.gz"
-    git_cmd = ["git", "archive", "--prefix", name_sha + "/", "--output", output_archive, "HEAD"]
-    log.debug("Generate source archive %r", git_cmd)
-    run(git_cmd, check=True, cwd=source_dir)
+
+    stash_cmd = ["git", "stash"]
+    log.debug("Stashing any changes to working repo %r", stash_cmd)
+    run(stash_cmd, check=True, cwd=source_dir)
+
+    mtime_cmd = ["git", "show", "-s", "--format=%cI"]
+    log.debug("Collecting timestamp of the commit at HEAD %r", mtime_cmd)
+    mtime_process = run(mtime_cmd, check=True, cwd=source_dir, capture_output=True, text=True)
+    mtime = mtime_process.stdout.strip()
+
+    ls_cmd = ["git", "ls-files", "--recurse-submodules"]
+    log.debug("Generate source repo file list %r", ls_cmd)
+    git_process = run(ls_cmd, check=True, cwd=source_dir, capture_output=True, text=True)
+
+    tar_cmd = [
+        "tar",
+        "caf",
+        output_archive,
+        "--mtime",
+        mtime,
+        "--transform",
+        f"s,^,{name_sha}/,",
+        "-T-",
+    ]
+    log.debug("Generate source archive %r", tar_cmd)
+    run(tar_cmd, input=git_process.stdout.encode("utf-8"), check=True, cwd=source_dir)
+
+    pop_cmd = ["git", "stash", "pop"]
+    log.debug("Popping any stashed changes to working repo %r", pop_cmd)
+    run(pop_cmd, cwd=source_dir)
+
     log.info("add source archive directory to sources for bsi: %s", source_archive_dir)
     sib_dirs.extra_src_dirs.append(source_archive_dir)
 
--- a/source-container-build/app/test_source_build.py
+++ b/source-container-build/app/test_source_build.py
@@ -1015,6 +1015,16 @@ class TestBuildProcess(unittest.TestCase):
                 completed_proc.stdout = "https://githost/org/app.git"
                 return completed_proc
 
+            if run_cmd == ["git", "ls-files"]:
+                completed_proc = Mock()
+                completed_proc.stdout = "file.txt"
+                return completed_proc
+
+            if run_cmd == ["git", "show"]:
+                completed_proc = Mock()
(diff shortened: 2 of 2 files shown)


cgwalters (approved):

source-container-build/app/source_build.py
```diff
@@ -391,9 +391,23 @@ def make_source_archive(
...
+
+    submodule_cmd = ["git", "submodule", "update", "--init", "--recursive"]
+    log.debug("Initializing submodule(s) %r", submodule_cmd)
+    run(submodule_cmd, check=True, cwd=source_dir)
+
+    ls_cmd = ["git", "ls-files", "--recurse-submodules"]
+    log.debug("Generate source repo file list %r", ls_cmd)
+    git_process = run(ls_cmd, check=True, cwd=source_dir, capture_output=True, text=True)
+
+    tar_cmd = ["tar", "caf", output_archive, "--transform", f"s,^,{name_sha}/,", "-T-"]
```
cgwalters, line 407: It's best practice to generate tarballs reproducibly; xref https://reproducible-builds.org/docs/env-variations/ and I recently did this in https://github.com/coreos/cargo-vendor-filterer/blob/0fa1892476943d1175089d89120eb694c0e3d11b/src/main.rs#L410 which note properly honors SOURCE_DATE_EPOCH but also falls back if that's not set to the toplevel git commit timestamp.

source-container-build/app/source_build.py
```diff
@@ -391,9 +391,23 @@ def make_source_archive(
     repo_info = get_repo_info(source_dir)
     name_sha = f"{repo_info['name']}-{repo_info['last_commit_sha']}"
     output_archive = f"{source_archive_dir}/{name_sha}.tar.gz"
-    git_cmd = ["git", "archive", "--prefix", name_sha + "/", "--output", output_archive, "HEAD"]
-    log.debug("Generate source archive %r", git_cmd)
-    run(git_cmd, check=True, cwd=source_dir)
+
+    restore_cmd = ["git", "restore", "."]
```
chmeliik, line 395: With this change, does the task now undo modifications in the source repo? AFAIK previously, the task would generate a clean archive but leave any modifications in the worktree untouched. IMO it'd be better to avoid this change in behavior unless its explicitly desired

ralphbean: Yeah, you're right. It's not explicitly desired to modify the worktree - but the intent is to generate the archive from the original committed content.

source-container-build/app/source_build.py
```diff
@@ -391,9 +391,23 @@ def make_source_archive(
...
+    log.debug("Restoring git repo %r", restore_cmd)
+    run(restore_cmd, check=True, cwd=source_dir)
+
+    submodule_cmd = ["git", "submodule", "update", "--init", "--recursive"]
```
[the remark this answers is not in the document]
ralphbean: Agreed - the clone task should have done this by now. I think this can be dropped.

source-container-build/app/source_build.py
chmeliik: > the intent is to generate the archive from the original committed content.

+1, we should do that. Maybe we can `git stash push` -> `git stash pop` the changes instead of discarding them?

source-container-build/app/source_build.py
ralphbean: Looks like the tar manual has guidance on this: https://www.gnu.org/software/tar/manual/html_section/Reproducibility.html

ralphbean: I think that's too much to take on in this code at this point in time.

chmeliik: I think setting at least the `--mtime` would be worth it, so that we get identical output when processing the same commit of a repo

```bash
$ function check_reproducibility() {
  git clone https://github.com/redhat-appstudio/build-tasks-dockerfiles /tmp/build-tasks-dockerfiles 2>/dev/null
  pushd /tmp/build-tasks-dockerfiles >/dev/null
  git ls-files | tar cza --mtime=$(git show -s --format=$cI) -T - | sha256sum
  popd >/dev/null
  rm -rf /tmp/build-tasks-dockerfiles
}

for i in {1..5}; do check_reproducibility; done
354e9e22a782f5773448355b44652b229b824936d7e1abbdedead7eb9b3330b0  -
354e9e22a782f5773448355b44652b229b824936d7e1abbdedead7eb9b3330b0  -
354e9e22a782f5773448355b44652b229b824936d7e1abbdedead7eb9b3330b0  -
354e9e22a782f5773448355b44652b229b824936d7e1abbdedead7eb9b3330b0  -
354e9e22a782f5773448355b44652b229b824936d7e1abbdedead7eb9b3330b0  -
```

chmeliik: git-archive does that as well: https://git-scm.com/docs/git-archive#_description

> When a tree ID is provided, the current time is used as the modification time of each file in the archive. On the other hand, when a commit ID or tag ID is provided, the commit time as recorded in the referenced commit object is used instead.

chmeliik (approved):
LGTM :+1:

source-container-build/app/source_build.py
tkdchen: Since source-build task runs in parallel with other tasks, `stash push -> stash pop` might cause conflict with potentially, although it is not recommended to make changes or write new artifacts inside the `source/`. `git worktree` can help without changing the `source/`.

source-container-build/app/source_build.py
```diff
@@ -391,9 +391,37 @@ def make_source_archive(
...
+    tar_cmd = [
+        "tar",
+        "caf",
+        output_archive,
+        "--mtime",
+        mtime,
+        "--transform",
+        f"s,^,{name_sha}/,",
+        "-T-",
+    ]
```
tkdchen, line 417: How about set `--owner=0 --group=0 --mode='a+rw'` explicitly?

PR #62, opened 2024-04-08
Reuse parent sources blobs for constructing source container

tkdchen:
STONEBLD-2189

This commit introduces a new method to include parent sources into the
source container. During the build process, the parent sources as well
as metadata are copied to the built source container directly without
extraction and being built with other sources together by BSI. As a
result,

* Parent sources are appended to the source container and are kept in
  the same order as in the original source container.
* Config data .rootfs.diff_ids and .history are updated into the source
  container config.
* The sources layers are updated into the source container manifest
  along with the config updates.

Major changes:

* Source container build process is changed. Now, it is:
  * download parent sources
  * build source container with application source and prefetched
    sources
  * copy parent sources into the built source container
  * push to the registry
* Parent sources are not extracted to local for merge.
* sourceImageBuildDirectories.rpm_dir is used for gathering prefetched
  SRPMs. No parent sources are extracted into the directory referenced
  by it.
* Both prepare_base_image_sources and extract_blob_member are useless
  and deleted.
* New Python classes are introduced for manipulating the OCI image,
  which are specific to this commit. For detailed information about
  these classes, like Blob, Config and Manifest, please refer to the
  docstring in the code.

This PR also includes a few related refactors, please check the individual commit.

### DIFF
--- a/source-container-build/app/source_build.py
+++ b/source-container-build/app/source_build.py
@@ -222,33 +222,23 @@ def extract_blob_member(
 
 
 def prepare_base_image_sources(
-    image: str, work_dir: str, sib_dirs: SourceImageBuildDirectories
+    source_image: str, work_dir: str, sib_dirs: SourceImageBuildDirectories
 ) -> bool:
     log = logging.getLogger("source-build.base-image-sources")
 
     base_image_sources_dir = create_dir(work_dir, "base_image_sources")
     base_sources_extraction_dir = create_dir(base_image_sources_dir, "extraction_dir")
 
-    source_image_name = resolve_source_image_by_version_release(image)
-
-    if not source_image_name:
-        logger.warning(
-            "The registry does not have corresponding source image %s", source_image_name
-        )
-        return False
-
     cmd = [
         "skopeo",
         "copy",
         "--retry-times",
         str(MAX_RETRIES),
-        f"docker://{source_image_name}",
+        f"docker://{source_image}",
         f"dir:{base_sources_extraction_dir}",
     ]
     log.info(
-        "Copy source image %s into directory %s",
-        source_image_name,
-        str(base_sources_extraction_dir),
+        "Copy source image %s into directory %s", source_image, str(base_sources_extraction_dir)
     )
     run(cmd, check=True)
 
@@ -580,8 +570,12 @@ def build(args) -> BuildResult:
 
         allowed = urlparse("docker://" + base_image).netloc in args.registry_allowlist
         if allowed:
-            prepared = prepare_base_image_sources(base_image, work_dir, sib_dirs)
-            build_result["base_image_source_included"] = prepared
+            source_image = resolve_source_image_by_version_release(base_image)
+            if source_image:
+                prepared = prepare_base_image_sources(source_image, work_dir, sib_dirs)
+                build_result["base_image_source_included"] = prepared
+            else:
+                logger.warning("Registry does not have source image %s", source_image)
         else:
             logger.info(
                 "Image %s does not come from supported allowed registry. "
--- a/source-container-build/app/source_build.py
+++ b/source-container-build/app/source_build.py
@@ -444,14 +444,9 @@ def make_source_archive(
     sib_dirs.extra_src_dirs.append(source_archive_dir)
 
 
-def build_and_push(
-    work_dir: str,
-    sib_dirs: SourceImageBuildDirectories,
-    bsi_script: str,
-    dest_images: list[str],
-    build_result: BuildResult,
-) -> None:
-    log = logging.getLogger("source-build.build-and-push")
+def build_source_image_in_local(
+    bsi_script: str, work_dir: str, sib_dirs: SourceImageBuildDirectories
+) -> str:
     bsi_build_base_dir = create_dir(work_dir, "bsi_build")
     image_output_dir = create_dir(work_dir, "bsi_output")
 
@@ -471,12 +466,15 @@ def build_and_push(
     if os.environ.get("BSI_DEBUG"):
         bsi_cmd.append("-D")
 
-    log.debug("build source image %r", bsi_cmd)
+    logger.debug("build source image %r", bsi_cmd)
     run(bsi_cmd, check=True)
+    return image_output_dir
 
-    # push to registry
+
+def push_to_registry(image_build_output_dir: str, dest_images: list[str]) -> str:
     fd, digest_file = tempfile.mkstemp()
     os.close(fd)
+    src = f"oci://{image_build_output_dir}:latest-source"
     for dest_image in dest_images:
         push_cmd = [
             "skopeo",
@@ -485,13 +483,13 @@ def build_and_push(
             digest_file,
             "--retry-times",
             str(MAX_RETRIES),
-            f"oci://{image_output_dir}:latest-source",
-            f"docker://{dest_image}",
+            src,
+            "docker://" + dest_image,
         ]
-        log.debug("push source image %r", push_cmd)
+        logger.debug("push source image %r", push_cmd)
         run(push_cmd, check=True)
     with open(digest_file, "r") as f:
-        build_result["image_digest"] = f.read().strip()
+        return f.read().strip()
 
 
 def generate_source_images(image: str) -> list[str]:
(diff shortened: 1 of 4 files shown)


tkdchen:
Test PR with e2e tests: https://github.com/redhat-appstudio/build-definitions/pull/917

source-container-build/app/source_build.py
```diff
@@ -528,6 +438,317 @@ def parse_image_name(image: str) -> tuple[str, str, str]:
...
+    parent_image = OCIImage(from_image)
+    local_build = OCIImage(to_image)
+
+    parent_image_manifest = parent_image.index.manifests()[0]
+    local_build_manifest = local_build.index.manifests()[0]
+    for layer in parent_image_manifest.layers:
+        copy_dest_path = Path(local_build.path, "blobs", *layer.descriptor["digest"].split(":"))
+        shutil.copyfile(layer.path, copy_dest_path)
+        logger.debug("copy layer %s to %s", layer.path, copy_dest_path)
+        local_build_manifest.append_layer(layer)
```
chmeliik, line 739: Doe we have any handling of duplicate layers? What happens if the local build has the same layer as the parent build?

source-container-build/app/source_build.py
```diff
@@ -528,6 +438,317 @@ def parse_image_name(image: str) -> tuple[str, str, str]:
...
+
+    parent_image_config = parent_image_manifest.config
+    local_build_config = local_build_manifest.config
+
+    local_build_config.diff_ids.extend(parent_image_config.diff_ids)
+    logger.debug("append diff_ids:\n%s", parent_image_config.diff_ids)
+    local_build_config.history.extend(parent_image_config.history)
+    logger.debug("append history:\n%s", parent_image_config.history)
```
chmeliik, line 747: Instead of *appending* the layers, diff_ids and history, can we *prepend* them? The parent image was built first, I would expect the parent layers to be first in the source container

source-container-build/app/source_build.py
```diff
@@ -528,6 +438,317 @@ def parse_image_name(image: str) -> tuple[str, str, str]:
...
+
+    @property
+    def layers(self) -> list[Layer]:
+        if self._layers is None:
+            self._layers = [Layer(self._layout, d) for d in self.to_python["layers"]]
+        return self._layers
+
+    def append_layer(self, layer: Layer) -> None:
+        layers: list[DescriptorT] = self.to_python["layers"]
+        layers.append(layer.descriptor)
```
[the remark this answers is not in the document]
tkdchen: This is intended. The modification to the layers will be available in the new manifest after calling `save`. The basic idea is, changes to a blob will be available in the new blob after save, and the current blob always references the current one.

source-container-build/app/source_build.py
tkdchen: So far, the parent sources are the ones introduced by a OS parent image, and both the prefetched SRPMs and extra sources have their own names and generated layers. The original solution with BSI didn't solve duplicates as well.

STONEBLD-1831 I'm working on will handle the potential duplicate layers between parent sources and prefetched extra sources, i.e. SRPMs and dependencies.

source-container-build/app/source_build.py
tkdchen: Any difference or special usage based on the order? My understanding is an image manifest just describes what are included in an image, seems the order of layers doesn't matter for a source container. BSI generates layers by finding out SRPMs depending on the default behavior of `find` and no guarantee to the order.

source-container-build/app/test_source_build.py
```diff
@@ -402,54 +373,71 @@ def test_build_with_extra_sources_only(self, run):
...
-
-        run.side_effect = run_side_effect
-        build_result: BuildResult = {}
-        self.sib_dirs.extra_src_dirs.append(mkdtemp())
-
-        source_build.build_and_push(
-            self.work_dir, self.sib_dirs, FAKE_BSI, [self.DEST_IMAGE], build_result
-        )
-
-        self.assertEqual(digest, build_result["image_digest"])
```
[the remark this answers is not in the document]
tkdchen: This line should cover that.

https://github.com/redhat-appstudio/build-tasks-dockerfiles/blob/main/source-container-build/app/test_source_build.py#L1161

source-container-build/app/source_build.py
```diff
@@ -528,6 +438,317 @@ def parse_image_name(image: str) -> tuple[str, str, str]:
...
 
+def download_parent_image_sources(source_image: str, work_dir: str) -> str:
+    """Download parent sources that stored in OCI image layout
+
+    :return: the directory holding the downloaded sources in the OCI image layout.
+    :rtype: str
+    """
+    sources_dir = create_dir(work_dir, "parent_image_sources")
+    logger.info("Copy source image %s into directory %s", source_image, sources_dir)
+    skopeo_copy(f"docker://{source_image}", f"oci:{sources_dir}", remove_signatures=True)
```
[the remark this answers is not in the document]
tkdchen: Updated

source-container-build/app/source_build.py
```diff
@@ -528,6 +438,317 @@ def parse_image_name(image: str) -> tuple[str, str, str]:
...
+        if not isinstance(that, self.__class__):
+            return False
+        this_d = self.descriptor
+        that_d = that.descriptor
+        return (
+            this_d["mediaType"] == that_d["mediaType"]
+            and this_d["digest"] == that_d["digest"]
+            and this_d["size"] == that_d["size"]
+            and this_d.get("annotations") == that_d.get("annotations")
+        )
```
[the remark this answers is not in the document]
tkdchen: Updated

source-container-build/app/source_build.py
```diff
@@ -528,6 +438,317 @@ def parse_image_name(image: str) -> tuple[str, str, str]:
...
+class JSONBlob(Blob):
+    """A blob whose content is encoded in JSON"""
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self._python_obj: dict | None = None
+
+    def save(self) -> Blob:
+        """Write JSON string in text mode"""
+        self.raw_content = json.dumps(self.to_python).encode("utf-8")
```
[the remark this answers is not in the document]
tkdchen: Updated

source-container-build/app/source_build.py
tkdchen: Update. PTAL.

source-container-build/app/source_build.py
```diff
@@ -528,6 +438,317 @@ def parse_image_name(image: str) -> tuple[str, str, str]:
...
+        if new_config != self.config:
+            self.to_python["config"] = new_config.descriptor
+        layer_descriptors = self.to_python["layers"]
+        for idx, layer in enumerate(self.layers):
+            if not layer.path.exists():
+                raise ValueError(f"layer {str(layer.path)} does not exist.")
+            new_layer = layer.save()
+            if new_layer != layer:
+                layer_descriptors[idx] = new_layer.descriptor
+                layer.delete()
```
chmeliik, line 650: > This is intended. The modification to the layers will be available in the new manifest after calling save. The basic idea is, changes to a blob will be available in the new blob after save, and the current blob always references the current one.

In that case, I probably don't understand why this is needed. If `self.layers` references the original layers, is this chunk of code needed? And now that we're prepending the parent layers instead of appending, does this work correctly?

source-container-build/app/source_build.py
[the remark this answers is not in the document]
chmeliik: Ok. It seems strange, but I won't dig deeper into this

source-container-build/app/source_build.py
```diff
@@ -710,35 +710,43 @@ def index(self) -> Index:
...
 
     parent_image_manifest = parent_image.index.manifests()[0]
     local_build_manifest = local_build.index.manifests()[0]
-    for layer in parent_image_manifest.layers:
+    for layer in reversed(parent_image_manifest.layers):
         copy_dest_path = Path(local_build.path, "blobs", *layer.descriptor["digest"].split(":"))
         shutil.copyfile(layer.path, copy_dest_path)
         logger.debug("copy layer %s to %s", layer.path, copy_dest_path)
-        local_build_manifest.append_layer(layer)
+        local_build_manifest.prepend_layer(layer)
```
chmeliik, line 734: Maybe it would be simpler to do things like this?

```python
for layer in local_build_manifest.layers:
    ...
    parent_image_manifest.append_layer(layer)

...
parent_image_manifest.save()
```

But that's a nitpick at this point, the current solution seems to work

chmeliik (commented):
One concern about a potential bug in https://github.com/redhat-appstudio/build-tasks-dockerfiles/pull/62#discussion_r1560703503, otherwise LGTM

Nice work :tada:

source-container-build/app/source_build.py
tkdchen: Good catch. Fixed.

tkdchen:
The `test_oci_image.py` was missing in the test list and not committed. This is also fixed.

PR #63, opened 2024-04-09
Fix SRPM collection

chmeliik:
Previously, the sib_dirs.rpms_dir would only get initialized in
prepare_base_image_sources.

While processing prefetched SRPMs, if base image sources were not
processed, the rpms_dir would still be unset. The SRPMs would get copied
into os.path.join("", ...) (so somewhere in the current directory) and
the script would ignore all of them, because the condition for including
the rpms_dir is that sib_dirs.rpms_dir must be set.

Fix by always initializing rpms_dir at the beginning and changing the
condition from "rpms_dir must be set" to "rpms_dir must not be empty".

### DIFF
--- a/source-container-build/app/source_build.py
+++ b/source-container-build/app/source_build.py
@@ -244,10 +244,6 @@ def prepare_base_image_sources(
     )
     run(cmd, check=True)
 
-    # bsi reads source RPMs from this directory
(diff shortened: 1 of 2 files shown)


tkdchen (commented):
#62 covers the rpm_dir issue already. Also no need to update `extract_blob_member` method, it is useless and removed in the #62

PR #66, opened 2024-04-22
Deduplicate sources

tkdchen:
STONEBLD-1831

If a parent image is built by Konflux, the sources introduced by it may duplicate with prefetched sources gathered during the local source build. Through the deduplicate process, comparing the parent sources and local build sources layer by layer. Once two layers have the same blob file and artifact name, they are treated as same, then the layer is removed from local source build. The same layer introduced by the parent will be merged into the source container image eventually.

Major changes:

* Add a new class BSILayer that wraps Layer to handle the layer archive generated by BuildSourceImage and implements a few methods for duplication check.
* Add a new method to dedpulicate the sources.
* Tests are updated with real tar archives.
* Shared code by tests are moved to test_utils.py specifically. This file intends to have shared code only and not include any test code.

Side effect:

* Same extra source tar archive names remain potentially. For instance, parent sources have extra-src-[0-3].tar and local source build has extra-src-[01].tar. When extra-src-1.tar duplicates the parent source extra-src-2.tar and is removed, then the final source image will have two layers including extra-src-0.tar.

  Whether this is a real problem in practice depends on how the sources are extracted. It should work well for extracting sources one after another. For the parallel extraction, separate working directory per layer should also work.


tkdchen:
Test PR: https://github.com/redhat-appstudio/build-definitions/pull/950

source-container-build/app/source_build.py
```diff
@@ -720,6 +765,126 @@ def index(self) -> Index:
...
+        return self._symlink_member  # type: ignore
+
+    @property
+    def blob_member(self) -> TarInfo:
+        """Return blob member
+
+        :return: a ``tarfile.TarInfo`` representing the blob member. None is
+            returned, if no such member is found from a layer.
+        """
+        return self._blob_member  # type: ignore
```
chmeliik, line 795: > None is returned, if no such member is found from a layer.

Should the return type be `TarInfo | None` then?

source-container-build/app/source_build.py
```diff
@@ -720,6 +765,126 @@ def index(self) -> Index:
...
+        return self.blob_member.name, artifact_name
+
+    def __hash__(self):
+        return hash(self.hash_key())
+
+    def __eq__(self, other: object) -> bool:
+        """Check if this layer contains same content of the other"""
+        if not isinstance(other, BSILayer):
+            return False
+        return hash(self) == hash(other)
```
chmeliik, line 885:
```suggestion
        return self.hash_key() == other.hash_key()
```
Otherwise different objects could be considered equal due to hash collisions

source-container-build/app/source_build.py
```diff
@@ -820,8 +1026,11 @@ def build(args) -> BuildResult:
 
     image_output_dir = build_source_image_in_local(args.bsi, work_dir, sib_dirs)
     if parent_sources_dir:
+        if build_result["dependencies_included"]:
+            deduplicate_sources(parent_sources_dir, image_output_dir)
```
chmeliik, line 1030: For my info, why do we de-duplicate only when dependencies are included?

chmeliik (commented):
Otherwise LGTM :+1:

source-container-build/app/source_build.py
tkdchen: The docstring is out-dated. The `_check` ensures there must be a blob member. This is also the reason why the comment is added.

source-container-build/app/source_build.py
tkdchen: My understanding is the duplication could happen to prefetched dependencies by Cachi2. The application application source and the sources introduced by the original parent images like OS images should not duplicate.

source-container-build/app/source_build.py
tkdchen: Updated.

tkdchen:
> > Same extra source tar archive names remain potentially. For instance, parent sources have extra-src-[0-3].tar and local source build has extra-src-[01].tar. When extra-src-1.tar duplicates the parent source extra-src-2.tar and is removed, then the final source image will have two layers including extra-src-0.tar.
> 
> Just noticed this. Is that not a problem? Using standard container filesystem extraction methods (e.g. `oc image extract`), the second `extra-src-0.tar` would overwrite the first one

It depends on how the sources are extracted. Like what I mentioned as well:

> Whether this is a real problem in practice depends on how the sources are extracted. It should work well for extracting sources one after another. For the parallel extraction, separate working directory per layer should also work.

Currently, marking it as a side-effect (or known issue) is based on the following a few considerations:

- It is not clear so far how the sources are used in practice, and there are a few ways to work around the side-effect with a slight change to user's extraction script.

- Since BuildSourceImage is used to construct the source container image, to adjust the counter for each extra source in local source build, the following steps have to be done:

    - extract the layer archive
    - update the counter for the archive member `extra-src-*.tar`
    - re-tar the layer archive with the same tar command BSI uses
    - remove the original layer
    - create a new layer from the new tar archive
    - compute `diff_id` and generate history for updating image config
    - insert the new layer

  Instead of taking such effort, probably the `extra-src-*.tar` file name can be in another form without counter. Then, next item ...

- The counter within `extra-src-*.tar` file name seems to be meaningless, but just for making unique file names. So, a better solution would be in the BuildSourceImage side by changing to an alternative file name pattern, like what trusted artifacts uses, e.g. may be `extra-src-sha256-checksum.tar`.

If you do think we should adjust the counter at current stage, I will add it.

chmeliik (approved):
LGTM :+1: 

But depending on what gets merged first between this and #73, one of the PRs will have to update the regex in `is_extra_src`

tkdchen:
> LGTM 👍
> 
> But depending on what gets merged first between this and #73, one of the PRs will have to update the regex in `is_extra_src`

Fixed. PTAL.

PR #73, opened 2024-04-24
Apply 0001-Use-extra-src-archive-checksum-in-filename.patch

tkdchen:
Related to STONEBLD-1831

Remove this patch when upgrade BuildSourceImage to a newer version that includes this change.


chmeliik (approved):
LGTM :+1: 

Just note that #66 will need corresponding changes in

```python
    def is_extra_src(member: TarInfo) -> bool:
        """Check if an archive member is a link of extra source archive
        Example arcname: ./extra_src_dir/extra-src-100.tar
        """
        dirname, basename = os.path.split(member.name)
        regex: Final = r"^extra-src-\d+\.tar$"
```

mkosiarc (approved):

PR #75, opened 2024-04-25
Resolve konflux source images

tkdchen:
STONEBLD-1831

This rework is for resolving source images built by Konflux. Since Konflux users are able to specify custom registry for pushing built images, the registry allow list does not make sense for resolving Konflux source images. So, as long as the source images cannot be resolved by version-release, source-build will try to resolve them as Konflux source image without consulting the allow list.

### DIFF
--- a/source-container-build/app/source_build.py
+++ b/source-container-build/app/source_build.py
@@ -445,6 +445,15 @@ def resolve_source_image_by_manifest(image: str) -> str | None:
         log.info("Source container image %s does not exist.", source_image)
 
 
+def resolve_source_image(binary_image: str, registries_allow_list: list[str]) -> str | None:
+    allowed = urlparse("docker://" + binary_image).netloc in registries_allow_list
+    if allowed:
+        return resolve_source_image_by_version_release(binary_image)
+    else:
+        logger.info("Image %s does not come from supported allowed registry.", binary_image)
+    return resolve_source_image_by_manifest(binary_image)
+
+
 def parse_image_name(image: str) -> tuple[str, str, str]:
     """Rough image name parser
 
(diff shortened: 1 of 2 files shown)


source-container-build/app/source_build.py
```diff
@@ -445,6 +445,19 @@ def resolve_source_image_by_manifest(image: str) -> str | None:
...
+    allowed = urlparse("docker://" + binary_image).netloc in registries_allow_list
+    if allowed:
+        return resolve_source_image_by_version_release(binary_image)
+    else:
+        logger.info(
+            "Image %s does not come from supported allowed registry. "
+            "Skip handling the sources for it.",
+            binary_image,
+        )
+    return resolve_source_image_by_manifest(binary_image)
```
chmeliik, line 458: Even if the image is in the "allowlist" (more like a list of registries where we expect to find `{version}-{release}.src` tags), we should still `resolve_source_image_by_manifest` for it.

Example: not all the images in `registry.access.redhat.com` use the `{version}-{release}.src` tagging scheme, some are built and released by Konflux and use the "cosign-style" tagging.

chmeliik: I know we discussed this offline, but can't quite remember what we agreed on. Could you summarize it here?

source-container-build/app/source_build.py
```diff
@@ -445,6 +445,19 @@ def resolve_source_image_by_manifest(image: str) -> str | None:
...
+def resolve_source_image(binary_image: str, registries_allow_list: list[str]) -> str | None:
+    allowed = urlparse("docker://" + binary_image).netloc in registries_allow_list
+    if allowed:
+        return resolve_source_image_by_version_release(binary_image)
+    else:
+        logger.info(
+            "Image %s does not come from supported allowed registry. "
+            "Skip handling the sources for it.",
+            binary_image,
+        )
```
chmeliik, line 457: The log message seems a bit misleading - the script says "Skip handling the source" and then immediately tries to resolve the source image by digest

tkdchen: Good catch :+1: PTAL

chmeliik (approved):
LGTM, though we will need to follow up on removing the allowlist

PR #79, opened 2024-04-28
Use match to clean up run_side_effects

tkdchen:
Meanwhile, this patch also makes them return real CompletedProcess instances instead of Mock objects.


mkosiarc (approved):

mkosiarc:
TIL about CompletedProcess

PR #82, opened 2024-05-03
fix: Provide insecure-policy option when copying source containers

ralphbean:
There's a bug in Red Hat production systems right now which is causing source containers to be published with no signatures:

* https://issues.redhat.com/browse/CLOUDDST-22731
* https://issues.redhat.com/browse/KFLUXBUGS-1267
* https://issues.redhat.com/browse/KFLUXBUGS-1268

We use those source containers as base images when constructing source containers for images built on Red Hat images. The absence of signatures is causing us to fail builds.

While those signatures are cleaned up, disable the policy check in skopeo to get builds moving again.

Consider, is this safe to apply? If we were faced with a similar situation where signatures were missing for binary images, we should not apply a workaround similar to this. Too risky. Source container images on the other hand are not a vector for attacking production systems.

### DIFF
--- a/source-container-build/app/source_build.py
+++ b/source-container-build/app/source_build.py
@@ -193,19 +193,26 @@ def fetch_image_manifest_digest(image: str) -> str:
 
 
 def skopeo_copy(
-    src: str, dest: str, digest_file: str = "", remove_signatures: bool = False
+    src: str,
+    dest: str,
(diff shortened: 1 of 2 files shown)


ralphbean:
Closing this. The source container signatures got fixed up.

PR #84, opened 2024-05-08
fix: fetch image manifest for generating source image

tkdchen:
When resolving parent source image with image manifest, the input image may include both tag and digest, which is generated by the build-container task. This fix ensures the image is cleaned by removing tag to make skopeo-inspect happy.


tkdchen:
Test PR: https://github.com/redhat-appstudio/build-definitions/pull/997

chmeliik (approved):
LGTM :+1:

rcerven (approved):

PR #101, opened 2024-06-18
fix: skip handling local parent image

tkdchen:
KFLUXBUGS-1293


mkosiarc (approved):

chmeliik (approved):

tkdchen:
Tested manually. source-build task is able to skip local image like `localhost/rhtap-final-image:none@sha256:10f29d...`.

PR #163, opened 2024-09-17
Added SPDX format support in sbom scripts [CLOUDDST-24250]

midnightercz:
Added support of SPDX format for helper scripts manipulating SBOMs [CLOUDDST-24250]
Related to ADR 0039-spdx-support


sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -92,6 +94,13 @@ def parse_args():
         description="Updates the sbom file with base images data based on the provided files"
     )
     parser.add_argument("--sbom", type=pathlib.Path, help="Path to the sbom file", required=True)
+    parser.add_argument(
+        "--sbom-type",
+        choices=["spdx", "cyclonedx"],
+        default="cyclonedx",
+        help="Type of the sbom file",
+        required=True,
+    )
```
chmeliik, line 103: Can we auto-detect this? Requiring the type to be explicitly specified will cause problems

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -127,10 +136,53 @@ def main():
...
+            sbom.update({"formulation": [{"components": base_images_sbom_components}]})
     else:
-        sbom.update({"formulation": [{"components": base_images_sbom_components}]})
+        packages = []
+        relationships = []
+        for component in base_images_sbom_components:
+            SPDXID = (
+                f"SPDXRef-{component['type']}-{component['name']}-"
+                + f"{hashlib.sha256(component['purl'].encode()).hexdigest()}"
+            )
```
chmeliik, line 151: What is the use case for the SPDXID?
- Is this something that will have to stay forever backwards compatible, because the IDs may be used as cross-SBOM references?
- Is there some reasoning behind the way the value is generated? Is it a good fit for the use cases?

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -127,10 +136,53 @@ def main():
...
+            packages.append(
+                {
+                    "SPDXID": SPDXID,
+                    "name": component["name"],
+                    "externalRefs": [
+                        {
+                            "referenceCategory": "PACKAGE-MANAGER",
```
chmeliik, line 158: Could you link some docs that explain the SPDX format? I didn't manage to find the referenceCategory anywhere

(Also wondering if PACKAGE-MANAGER is the right category, but I wasn't able to find the list of categories)

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -127,10 +136,53 @@ def main():
...
+                    "annotations": [
+                        {
+                            "annotator": "konflux",
+                            "annotationDate": datetime.datetime.now().isoformat(),
+                            "annotationType": "OTHER",
+                            "comment": json.dumps(
+                                {"name": property["name"], "value": property["value"]},
+                                separators=(",", ":"),
+                            ),
+                        }
```
chmeliik, line 172: What is the intended interface here? Are we stating that consumers of Konflux-made SPDX SBOMs should expect all the `annotations` with `annotator=konflux` to be in the form `{"name": "...", "value":  "..."}`?

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -127,10 +136,53 @@ def main():
...
+                        {
+                            "referenceCategory": "PACKAGE-MANAGER",
+                            "referenceType": "purl",
+                            "referenceLocator": component["purl"],
+                        }
+                    ],
+                    "annotations": [
+                        {
+                            "annotator": "konflux",
+                            "annotationDate": datetime.datetime.now().isoformat(),
```
chmeliik, line 166: Let's get the datetime once to avoid each annotation having a different datetime

chmeliik (commented):
Thank for the contribution!

Could you give the PR a more thorough description and link the relevant issues, docs etc.? There's a lot of guesswork for us here.

(Reviewed only `base_images_sbom_script.py` so far)

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -127,10 +136,53 @@ def main():
...
+                        }
+                        for property in component["properties"]
+                    ],
+                }
+            )
+            relationships.append(
+                {
+                    "spdxElementId": sbom["SPDXID"],
+                    "relatedSpdxElement": SPDXID,
+                    "relationshipType": "BUILD_TOOL_OF",
```
chmeliik, line 181: What is `BUILD_TOOL_OF` supposed to represent? A tool used to build the element referenced by `spdxElementId`, i.e. SBOM itself? That would not be the right relationship. The builder images play no part in generating the SBOM.

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
midnightercz: I guess we can, but what if you want to for example work with cyclonedx but you receive spdx. Idea is that we switch whole pipeline to spdx with single parameter, so unexpected format should produce a failure

chmeliik: > Idea is that we switch whole pipeline to spdx with single parameter

That's not really possible. Konflux proposes the initial pipeline to the user's repo in a PR, but after that, the pipeline is fully in the user's control. "Switch to spdx with a single parameter" would be more like "every Konflux user needs to set this parameter for all the tasks in all their pipelines."

Think of it this way - the main unit that Konflux produces is the Task, not the Pipeline. Interoperability and backwards compatibility needs to be handled on the Task level

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
midnightercz: It's unique identifier which can be used externally, so yes, it's something permanent. 

> Is there some reasoning behind the way the value is generated? Is it a good fit for the use cases?
syft generates it in similar way, but it uses some go hasher which calculates hash of whole structure

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
midnightercz: Cachi2 uses some cyclonedx properties, however spdx format (v2.3) doesn't have any field which we could use for that, so we decided to use package annotations for that.

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
chmeliik: > It's unique identifier which can be used externally, so yes, it's something permanent.

Ack, please add a comment about that

> syft generates it in similar way, but it uses some go hasher which calculates hash of whole structure

Hmm, interesting. If two elements have the same purl but some other property is different, should they be considered the same element? I guess Syft's answer is No, but maybe it should be Yes.

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
chmeliik: Right, I remember now. The format looks weird, but I don't really have a better idea. It will just need to be documented somewhere.

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
chmeliik: Then I think we're not using it correctly. `SPDXID` refers to a base image, `sbom["SPDXID"]` refers to the SBOM. The base image is not used to build the SBOM.

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
midnightercz: I was told users basically creates custom "PipelineRuns" but the pipelines and task definitions are take from the main repo. So idea is to have default set to cyclonedx and after we're ready to switch to spdx we can change the default to spdx. Users don't even needs to know there's some parametr which controls sbom format

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
midnightercz: yeah, I would say syft would generate different SPDXID, but question is if there isn't some deduplication mechanism which it uses internally.

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
midnightercz: I guess cachi2 would be the best place to document this.

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
chmeliik: Yes, when a user onboards to Konflux, their pipeline is generated based on the state of build-definition at the time. But when the pipeline definition in build-definition changes, the pipeline in users' repos does not.

So the switch to SPDX will require a mass migration

chmeliik: But since we'll likely need to do that mass migration no matter what, maybe requiring a param is not so bad. It is still worse than making the buildah task work regardless of input SBOM format, but if it would add that much extra complexity, we can rely on a param

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
chmeliik: Might be, but this also applies cross-SBOM. If two different SBOMs have a component with the same purl, should it have the same SPDXID? Probably yes.

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
chmeliik: For the cachi2 annotations, I would agree. I guess those use `"annotator": "cachi2"`?

For the Konflux annotations, probably somewhere in https://github.com/konflux-ci/docs

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
chmeliik: You changed this in a later commit, but I don't understand the change. Could you explain which two elements we are now marking as related?

chmeliik: (also, please update these threads when you make relevant changes :pray: )

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
midnightercz: I think if you refer to package in another document it has to be referenced by the spdxid of the document and its own spdxid.
"
7.2 Package SPDX identifier field
7.2.1 Description

Uniquely identify any element in an SPDX document which may be referenced by other elements. These may be referenced internally and externally with the addition of the SPDX document identifier.
"

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
midnightercz: I added extra comments to the code to explain relationship generation

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
mkosiarc: auto-detection would be great, but like @chmeliik  is saying maybe having the parameter would be sufficient solution here. Should we discuss this in the build team @chmeliik so we can resolve this?

chmeliik: I'd like input from @midnightercz on the complexity of making it smarter. It feels trivial.

mkosiarc:
If the newly added scripts are to be used in the buildah task in konflux, they need to be added to the Dockerfile as well - https://github.com/konflux-ci/build-tasks-dockerfiles/blob/main/sbom-utility-scripts/Dockerfile

This means it will be part of the image that is used in the individual steps that are working with sbom in the buildah task, e.g. https://github.com/konflux-ci/build-tasks-dockerfiles/blob/main/sbom-utility-scripts/Dockerfile

mkosiarc:
Is this work tracked internally somewhere?

mkosiarc:
Seems that the CI did not run automatically because you are not a maintainer. I had to approve it and it seems to run now. Sorry about that.

The new testcase is failing:
```
        else:
>           root_element1, map1, inverse_map1 = map_relationships(sbom['relationships'])
E           KeyError: 'relationships'
base_images_sbom_script.py:165: KeyError
=========================== short test summary info ============================
FAILED test_base_images_sbom_script.py::test_main_input_sbom_spdx_minimal - K...
```

You can also test this locally with `tox -e test`

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
chmeliik: Ack, makes sense.

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -127,10 +160,90 @@ def main():
...
+
+        # Try to calculate middle element based on the relationships maps.
+        # SPDX has usually root element which contains a wrapper element which then contains
+        # all of the other elements
+        middle_element1 = None
+        for r, contains in map1.items():
+            if contains and inverse_map1.get(r) == root_element1:
+                middle_element1 = r
+        # if not middle_element1:
+        #    middle_element1 = root_element1
```
mkosiarc, line 182: this is probably left by mistake?

mkosiarc:
@chmeliik does @midnightercz need to be added to the rhtap-build namespace, so that the pipeline can be triggered from this PR?

chmeliik:
just needs /ok-to-test no?

mkosiarc:
> > Is this work tracked internally somewhere?
> 
> It's tracked in CLOUDDST-24250, I created the issue after I started working on this, so that's why it's not in the commit mesage

You could always edit the commit message and make life easier for everybody in the future, but now it is at least mentioned in the pull request.

mmorhun:
@midnightercz could you rebase the PR?

tkdchen (commented):
Please describe the problem, which this PR addresses, in detail in the commit message of 1b01170836b82b49b75d501648cd78c119e83697. The information may be about what is missing by current implementation, what's the purpose of adding SPDX format support, etc.

midnightercz:
> Please describe the problem, which this PR addresses, in detail in the commit message of [1b01170](https://github.com/konflux-ci/build-tasks-dockerfiles/commit/1b01170836b82b49b75d501648cd78c119e83697). The information may be about what is missing by current implementation, what's the purpose of adding SPDX format support, etc.

I added information about ADR describing the problem

chmeliik:
> What can be done to get this merged? The EC team is awaiting this work to be done before proceeding with the full SPDX support.

Currently the SPDX feature is in design phase https://github.com/konflux-ci/architecture/pull/213

zregvart:
> Also, question: Is `merge_syft_sboms_spdx.py` used anywhere? It looks like it's not used in build-definitions repo. Is it possible it's deprecated script?

I've found no use of it anywhere

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -87,11 +89,26 @@ def get_base_images_sbom_components(base_images_digests, is_last_from_scratch):
...
         description="Updates the sbom file with base images data based on the provided files"
     )
     parser.add_argument("--sbom", type=pathlib.Path, help="Path to the sbom file", required=True)
+    parser.add_argument(
+        "--sbom-type",
+        choices=["spdx", "cyclonedx"],
+        default="cyclonedx",
+        help="Type of the sbom file",
+        required=True,
+    )
```
chmeliik, line 111: This is unused now

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -127,10 +168,96 @@ def main():
...
+        # Try to calculate root package represeting the container image or directory, which was
+        # used to build the SBOM, based on the relationships maps.
+        # SPDX has relationsship ROOT-ID DESCRIBES MIDDLE-ID which express the fact the SBOM documents
+        # describes container image or directory represented by MIDDLE-ID package.
+        root_package1 = None
+        for r, contains in map1.items():
+            # root package is the one which contains another elements and is in relationship with
+            # the document element where it stand as relatedSpdxElement
+            if contains and inverse_map1.get(r) == root_element1:
+                root_package1 = r
```
chmeliik, line 191: > SPDX has relationsship ROOT-ID DESCRIBES MIDDLE-ID

The goal is straightforward, but the code is not. Could you simplify the code to just find the `ROOT-ID DESCRIBES MIDDLE-ID` relationship?

chmeliik: In [add_image_reference.py](https://github.com/konflux-ci/build-tasks-dockerfiles/blob/main/sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py), we also have handling for the case when there are multiple `ROOT-ID DESCRIBES *` relationships. Should we:

* raise an error here if that happens, always call this script _after_ `add_image_reference.py`
* have similar handling here as `add_image_reference.py`
* other options?

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -127,10 +168,96 @@ def main():
...
+                    "relatedSpdxElement": "SPDXRef-DocumentRoot-Unknown-",
+                    "relationshipType": "DESCRIBES",
+                }
+            )
+
+        annotation_date = datetime.datetime.now().isoformat()
+        for component in base_images_sbom_components:
+            # Calculate unique identifier SPDXID based on the component name and purl
+            SPDXID = (
+                f"SPDXRef-{component['type']}-{component['name']}-"
```
chmeliik, line 215: Other scripts in this repo use `SPDXRef-image`, whereas I think this would result in `SPDXRef-container`

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -127,10 +168,96 @@ def main():
...
+                            "referenceCategory": "PACKAGE-MANAGER",
+                            "referenceType": "purl",
+                            "referenceLocator": component["purl"],
+                        }
+                    ],
+                    # Annotations are used to provide cyclonedx custom properties
+                    # as json string
+                    "annotations": [
+                        {
+                            "annotator": "Tool:konflux:jsonencoded",
```
chmeliik, line 236:
```suggestion
                            "annotator": "Tool: konflux:jsonencoded",
```

sbom-utility-scripts/scripts/base-images-sbom-script/app/test_base_images_sbom_script.py
```diff
@@ -516,6 +517,131 @@ def test_main_input_sbom_does_not_contain_formulation(tmp_path, mocker):
     assert expected_output["formulation"] == sbom["formulation"]
 
 
+@pytest.fixture
+def isodate():
+    with patch("datetime.datetime") as mock_datetime:
+        mock_datetime.now.return_value.isoformat.return_value = "2021-07-01T00:00:00Z"
```
chmeliik, line 523: This doesn't match the actual output format of `isoformat()`

sbom-utility-scripts/scripts/base-images-sbom-script/app/test_base_images_sbom_script.py
```diff
@@ -516,6 +517,131 @@ def test_main_input_sbom_does_not_contain_formulation(tmp_path, mocker):
...
+    sbom_file.write_text(
+        """{
+    "SPDXID": "SPDXRef-Document",
+    "spdxVersion": "SPDX-2.3",
+    "name": "MyProject",
+    "documentNamespace": "http://example.com/uid-1234",
+    "packages": [],
+    "relationships": []
+    }"""
+    )
```
chmeliik, line 542: Do we have any non-minimal tests?

sbom-utility-scripts/scripts/base-images-sbom-script/app/test_base_images_sbom_script.py
```diff
@@ -516,6 +517,131 @@ def test_main_input_sbom_does_not_contain_formulation(tmp_path, mocker):
...
+                "relationshipType": "BUILD_TOOL_OF",
+                "spdxElementId": "SPDXRef-container-quay.io/mkosiarc_rhtap/"
+                "single-container-app-9520a72cbb69edfca5cac88ea2a9e0e09142ec934952b9420d686e77765f002c",
+            },
+            {
+                "spdxElementId": "SPDXRef-container-registry.access.redhat.com/"
+                "ubi8/ubi-0f22256f634f8205fbd9c438c387ccf2d4859250e04104571c93fdb89a62bae1",
+                "relationshipType": "BUILD_TOOL_OF",
+                "relatedSpdxElement": "SPDXRef-DocumentRoot-Unknown-",
+            },
```
chmeliik, line 634: Could you keep `spdxElementId` and `relatedSpdxElement` in a consistent order, ideally `spdxElementId` first?

At first, I thought "Wait, how the heck does `SPDXRef-DocumentRoot-Unknown-` describe the `SPDXRef-Document`" until I noticed the order was inconsistent

sbom-utility-scripts/scripts/create_purl_sbom_spdx.py
chmeliik, line 1: You can drop this, we don't create purl SBOMs anymore after https://github.com/konflux-ci/build-tasks-dockerfiles/pull/193

sbom-utility-scripts/scripts/merge-cachi2-sboms-script/merge_cachi2_sboms.py
```diff
@@ -148,6 +299,199 @@ def _merge_tools_metadata(syft_sbom: dict[Any, Any], cachi2_sbom: dict[Any, Any]
...
+    root_package2 = calculate_root_package(root_element2, map2, inverse_map2)
+
+    for relation in relationships2:
+        _relation = relation.copy()
+
+        # If relations is Root decribes middle element, skip it
+        if (
+            _relation["relatedSpdxElement"] == root_package2
+            and _relation["spdxElementId"] == root_element2
+            and _relation["relationshipType"] == "DESCRIBES"
```
chmeliik, line 440: I tried giving the variables less confusing names and the code started to make a bit more sense

```suggestion
            _relation["relatedSpdxElement"] == root2
            and _relation["spdxElementId"] == doc2
            and _relation["relationshipType"] == "DESCRIBES"
```

(inspired by https://github.com/midnightercz/konflux-ci-architecture/blob/spdx-support/ADR/0044-spdx-support.md#relationships)

sbom-utility-scripts/scripts/merge-cachi2-sboms-script/merge_cachi2_sboms.py
```diff
@@ -148,6 +299,199 @@ def _merge_tools_metadata(syft_sbom: dict[Any, Any], cachi2_sbom: dict[Any, Any]
...
+        if _relation["spdxElementId"] == root_package2:
+            _relation["spdxElementId"] = root_package1
+        if _relation["relatedSpdxElement"] == root_package2:
+            _relation["relatedSpdxElement"] = root_package1
+
+        # include only relations to packages which exists in merged packages.
+        if _relation["relatedSpdxElement"] in package_ids:
+            relationships.append(_relation)
+        elif _relation["spdxElementId"] in package_ids:
+            relationships.append(_relation)
```
chmeliik, line 458:
```suggestion
        if (
            _relation["relatedSpdxElement"] in package_ids
            or _relation["spdxElementId"] in package_ids
        ):
            relationships.append(_relation)

```

sbom-utility-scripts/scripts/merge-cachi2-sboms-script/merge_cachi2_sboms.py
```diff
@@ -148,6 +299,199 @@ def _merge_tools_metadata(syft_sbom: dict[Any, Any], cachi2_sbom: dict[Any, Any]
...
+
+    for relation in relationships1:
+        _relation = relation.copy()
+        if relation["relatedSpdxElement"] in package_ids:
+            relationships.append(_relation)
```
chmeliik, line 463: Why do we only look at `relatedSpdxElement` for document 1 when we look at both `spdxElementId` and `relatedSpdxElement` for document 2?

sbom-utility-scripts/scripts/merge-cachi2-sboms-script/merge_cachi2_sboms.py
```diff
@@ -148,6 +299,199 @@ def _merge_tools_metadata(syft_sbom: dict[Any, Any], cachi2_sbom: dict[Any, Any]
...
+            "annotationType": annotation[3],
+        }
+        for annotation in annotations
+    ]
+
+
+def merge_relationships(relationships1, relationships2, packages):
+    """Merge SPDX relationships."""
+
+    def map_relationships(relationships):
```
chmeliik, line 392: The method of looking for the Document SPDXID and the root SPDXID is similarly overcomplicated and with similar deficiencies as https://github.com/konflux-ci/build-tasks-dockerfiles/pull/163#discussion_r1863588887

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -127,10 +168,96 @@ def main():
...
+                    "relationshipType": "DESCRIBES",
+                }
+            )
+
+        annotation_date = datetime.datetime.now().isoformat()
+        for component in base_images_sbom_components:
+            # Calculate unique identifier SPDXID based on the component name and purl
+            SPDXID = (
+                f"SPDXRef-{component['type']}-{component['name']}-"
+                + f"{hashlib.sha256(component['purl'].encode()).hexdigest()}"
```
tkdchen, line 216: Please extract it to a separate method and document it in docstring. Especially, describing the format of the unique SPDXID, and pointing to relative SPDX specification if there is.

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -127,10 +168,96 @@ def main():
...
+        root_package1 = None
+        for r, contains in map1.items():
+            # root package is the one which contains another elements and is in relationship with
+            # the document element where it stand as relatedSpdxElement
+            if contains and inverse_map1.get(r) == root_element1:
+                root_package1 = r
+        # If not root package is found then create one with ID "Uknown" as source for the SBOM
+        # is not known.
+        if not root_package1:
+            root_package1 = "SPDXRef-DocumentRoot-Unknown-"
```
tkdchen, line 195: Why does `root_package1` have this string value and why does it end with `-`? There seems no code appending other string behind the tailing `-`.

sbom-utility-scripts/scripts/base-images-sbom-script/app/test_base_images_sbom_script.py
midnightercz: Yeah, it is not. But follows the specification. Should we rename it? I chose isodate as it's format which resemble is the most.

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
chmeliik: We now have and ADR that mentions the annotations, should be good enough

sbom-utility-scripts/scripts/base-images-sbom-script/app/test_base_images_sbom_script.py
chmeliik: IMO we should the format that matches the specification (mainly in the production code, not just in test mocks)

sbom-utility-scripts/scripts/merge-cachi2-sboms-script/merge_cachi2_sboms.py
```diff
@@ -148,6 +299,199 @@ def _merge_tools_metadata(syft_sbom: dict[Any, Any], cachi2_sbom: dict[Any, Any]
...
+                cpackage = cachi2_packages_map[get_package_key(p)]
+                cpackage["externalRefs"] = sorted(
+                    merge_external_refs(cpackage.get("externalRefs", []), p.get("externalRefs", [])),
+                    key=lambda x: (
+                        x["referenceCategory"],
+                        x["referenceType"],
+                        x["referenceLocator"],
+                    ),
+                )
+                cpackage["annotations"] = merge_annotations(cpackage.get("annotations", []), p.get("annotations", []))
```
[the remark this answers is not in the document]
midnightercz: for cyclonedx if the component is duplicated it's based on purl. That means the component is duplicate for sure. With SPDX considering possibility of multiple purls, is_duplicate_package returns True even on partial match of purls. Therefore considered removal of duplicated packages as possible  information loss and that's why attributes of the duplicate package are merged with existing one instead of discarding whole package

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
midnightercz: I think it makes sense to raise exception in such case as many other scripts depend on the same condition

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
midnightercz: Yeah, that's how syft does that. Probably it use naming convention {name}-{version}. I think it can be removed.

sbom-utility-scripts/scripts/merge-cachi2-sboms-script/merge_cachi2_sboms.py
midnightercz: Because in the first doc spdxElementId can be root element and in that case even relationships pointing to removed packages would be included

sbom-utility-scripts/scripts/merge-cachi2-sboms-script/merge_cachi2_sboms.py
```diff
@@ -114,6 +225,46 @@ def component_is_duplicated(component: dict[str, Any]) -> bool:
...
+    def package_is_duplicated(package: dict[str, Any]) -> bool:
+        keys = _unique_keys_syft_spdx(package)
+
+        return (
+            _is_syft_local_golang_package(package)
+            or is_duplicate_non_registry_package(package)
+            or set(keys) & set(cachi2_indexed_packages.keys())
+        )
+
+    return package_is_duplicated
```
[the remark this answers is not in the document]
midnightercz: code duplication is intentional. Once we remove cyclonedx support, we can simply remove all non spdx related functions. If we enforce to support only one purl per package then logic could be simplified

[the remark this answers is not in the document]
chmeliik: > If we enforce to support only one purl per package then logic could be simplified

yes please :pray:

chmeliik:
Everything that was needed has now been implemented

PR #165, opened 2024-09-25
Feat: Add script for image index SBOM creation.

BorekZnovustvoritel:
Ticket for tracking: [ISV-5220](https://issues.redhat.com/browse/ISV-5220).

This task is a part of an effort to create SBOMs for index images (manifest lists). The work breakdown can be found in the links of [this ticket](https://issues.redhat.com/browse/ISV-5162), which is a part of the epic [Manifest list SBOM](https://issues.redhat.com/browse/CLOUDWF-10585).

After this PR is merged, another PR should follow to [Konflux task `build-image-index`](https://github.com/konflux-ci/build-definitions/tree/main/task/build-image-index), which should call this script. The current consent is that the Tekton task should call `buildah manifest inspect` on the newly created index, then provide the output to this script along with the index url and digest.

This script should output a valid SPDX 2.3 SBOM.


sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
```diff
@@ -0,0 +1,212 @@
...
+import json
+from datetime import datetime, timezone
+from typing import Optional
+from dataclasses import dataclass, field
+from uuid import uuid4
+
+from packageurl import PackageURL
+
+
+SUPPORTED_ARCHITECTURES = ["amd64", "arm64", "s390x", "ppc64le"]
```
Allda, line 12: This script should be architecture agnostic. Ideally, there should not be any hardcoded value for an architecture. The pipeline itself defines which architectures will be built so this script should just take the output from the pipeline task and assemble the sbom. In case a new arch is added or anything related to the arch is modified this script becomes a blocker for the rest of the pipeline.

BorekZnovustvoritel: Thanks for pointing that out. I have reworked the script so now it works even if no architecture of the child image is specified and no value is hardcoded. It just produces one less PURL if that's the case. If you have further points to improve this, be sure to let me know :saluting_face:

BorekZnovustvoritel: I have reworked it again, now it should work with the output of command `buildah manifest inspect` instead of the long list of arguments as it was previously.

BorekZnovustvoritel:
The script should be ready for a review. If you have any points of improvement, I'll be happy to include them. Not sure who to tag as a repository owner...
@chmeliik @mkosiarc

chmeliik:
@BorekZnovustvoritel thanks for contributing!

Some more context in the PR description/commit messages would go a long way to make this easier for us to review. What is this script for, how will it be used, etc. If there's a design for it somewhere a link would be appreciated

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
```diff
@@ -0,0 +1,196 @@
...
+        )
+        return ans
+
+    def propose_spdx_id(self) -> str:
+        return f"SPDXRef-{self.digest_hex_val}"
+
+
+def create_package(image: Image, spdxid: Optional[str] = None, image_index_digest: Optional[str] = None) -> dict:
+    return {
+        "SPDXID": image.propose_spdx_id() if not spdxid else spdxid,
```
chmeliik, line 72: The conversation [here](https://github.com/konflux-ci/build-tasks-dockerfiles/pull/163#discussion_r1766359666) has led me to believe that the format of SPDXID is a decision with long-term impact. In every PR that adds SPDX support, there's a different approach taken for the SPDXID
* this one
* https://github.com/konflux-ci/build-tasks-dockerfiles/pull/163
* https://github.com/containerbuildsystem/cachi2/pull/608

It would be great if you could sync with @midnightercz and the stakeholders (Red Hat prodsec?) on what the SPDXID should be

BorekZnovustvoritel:
> @BorekZnovustvoritel thanks for contributing!
> 
> Some more context in the PR description/commit messages would go a long way to make this easier for us to review. What is this script for, how will it be used, etc. If there's a design for it somewhere a link would be appreciated

Sorry for the lack of clarity, I have updated the PR description. If something is still missing, let me know and I will add additional info.

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
BorekZnovustvoritel: I have tried to search for guidelines from ProdSec and I have found [this](https://github.com/RedHatProductSecurity/security-data-guidelines/blob/main/docs/sbom.md#packages-and-relationships). It states that the `SPDXID` must be unique *within* the SBOM, but I agree that we should double-check with them so we don't prepare any traps for ourselves in the future.

Allda (approved):

BorekZnovustvoritel:
Hello, any updates on the review? I have synced with [midnightercz](https://github.com/midnightercz) and implemented his SPDXID creation mechanism as well.

chmeliik:
Sorry, I'm on konflux-user support rotation. I've had a total of about 1 hour of time between support threads this week (╯°□°）╯︵ ┻━┻

Will get to this eventually. I still have to read through the design to understand how this will be used.

mkosiarc:
The code looks ok to me, the tests are passing, so I will provide an ack. I don't have the time to go through the design, so I will just assume that this phase was reviewed and everything worked out. 

But please update one more thing - https://github.com/konflux-ci/build-tasks-dockerfiles/blob/main/.github/workflows/build-sbom-utility-scripts-image.yml#L34

Update the file so it will run a tox check on your added script as well. This will make sure that the tests are passing.

BorekZnovustvoritel:
Thank you for the feedback, workflow step added.

mkosiarc (approved):

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
```diff
@@ -0,0 +1,198 @@
...
+        purl_hex_digest = hashlib.sha256(self.purls()[0].encode()).hexdigest()
+        return f"SPDXRef-Image-{self.name}-{purl_hex_digest}"
+
+
+def create_package(image: Image, spdxid: Optional[str] = None, image_index_digest: Optional[str] = None) -> dict:
+    return {
+        "SPDXID": image.propose_spdx_id() if not spdxid else spdxid,
+        "name": image.name if not image.arch else f"{image.name}_{image.arch}",
+        "versionInfo": image.tag,
+        "supplier": "Organization: Red Hat",
```
chmeliik, line 77: How much do we need the `supplier` information?

We shouldn't hardcode `Red Hat`, we can't assume only Red Hat is using Konflux. In fact, I believe Fedora is setting up / has set up their Konflux instance as well.

Can we leave it as NOASSERTION or remove it altogether?

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
```diff
@@ -0,0 +1,198 @@
...
+            tag=image_index_obj.tag,
+            repository=image_index_obj.repository,
+        )
+        packages.append(create_package(arch_image, image_index_digest=image_index_obj.digest))
+        relationships.append(get_relationship(arch_image.propose_spdx_id(), "SPDXRef-image-index"))
+
+    sbom = {
+        "spdxVersion": "SPDX-2.3",
+        "dataLicense": "CC0-1.0",
+        "documentNamespace": f"https://redhat.com/spdxdocs/{sbom_name}-{uuid4()}",
```
chmeliik, line 142: Similar concern about hardcoding Red Hat here

chmeliik (commented):
Nice code! :taco: 

Just a few functional concerns

sbom-utility-scripts/Dockerfile
```diff
@@ -7,5 +7,7 @@ COPY scripts/merge-cachi2-sboms-script/merge_cachi2_sboms.py /scripts
 COPY scripts/create_purl_sbom.py /scripts
 COPY scripts/base-images-sbom-script/app/base_images_sbom_script.py /scripts
 COPY scripts/base-images-sbom-script/app/requirements.txt /scripts
+COPY scripts/index-image-sbom-script/index_image_sbom_script.py /scripts
 
+# Same requirements for base-images-sbom-script and index-image-sbom-script
 RUN pip3 install -r requirements.txt
```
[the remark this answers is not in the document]
chmeliik: > very nasty surprises

For example, the index-image requirements.txt currently specifies a higher version of `packageurl-python` than the base-images requirements.txt. That means we're unit testing with a different version that what we'll be running in prod, which is a bit scary

chmeliik:
My understanding of the overall feature:

This script will be used in the `build-image-index` task.

The content of the SBOM as generated by this script will be the entire SBOM - we will not be merging the arch-specific SBOMs into it in any way.

We will push the SBOM to the registry and associate it with the index image the same way that the arch-specific SBOMs are associated with the arch-specific manifests.

Some system will then be able to associate the index-image SBOM with the arch-specific SBOMs based on... something. This is the part that I'm missing - what is the "something" that will allow a consumer to associate the SBOMs?

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
BorekZnovustvoritel: This is taken from ProdSec guidelines, as seen [here](https://redhatproductsecurity.github.io/security-data-guidelines/sbom/#packages-and-relationships). However I can see how this can become a problem. I will change it to `NOASSERTION`.

BorekZnovustvoritel:
Thank you for the review. I will rework this PR to resolve all comments. Your understanding of the overall feature is correct (or at least we understand it in the same way :grinning:). The linking to the outside is supposed to be represented by the `externalRefs` field, where PURLs are specified, as represented [here](https://github.com/RedHatProductSecurity/security-data-guidelines/blob/main/sbom/examples/container_image/build/ubi9-micro-container-9.4-6.1716471860.spdx.json).

If you prefer some other method of referencing, be sure to comment on it, however that would require bigger changes to the [Build Image Index task](https://github.com/konflux-ci/build-definitions/tree/main/task/build-image-index) as there is not enough information for other SBOM linking right now.

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
BorekZnovustvoritel: Changed `redhat.com` to `konflux-ci.dev`.

sbom-utility-scripts/Dockerfile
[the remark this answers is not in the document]
BorekZnovustvoritel: Thanks for pointing that out. Changed according to your proposal and unified the package versions of the 2 scripts (pip installation fails otherwise).

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
chmeliik: Oh, it's just a "made up" URL. This works :smile:

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
```diff
@@ -0,0 +1,198 @@
...
+        ans.append(
+            PackageURL(
+                type="oci", name=self.name, version=self.digest, qualifiers={"repository_url": self.repository}
+            ).to_string()
+        )
+        return ans
+
+    def propose_spdx_id(self) -> str:
+        purl_hex_digest = hashlib.sha256(self.purls()[0].encode()).hexdigest()
+        return f"SPDXRef-Image-{self.name}-{purl_hex_digest}"
```
chmeliik, line 69: nitpick: if we want to include the type in the SPDXID, let's be consistent about lowercase/uppercase (`Image` vs. `image-index`)

Alternatively, maybe we can just stick to `SPDXRef-{name}-{purl_digest}` (for everything, the image index and the images). It looks like your PR will be going in first, so we can set the precedent here :slightly_smiling_face:

chmeliik (commented):
LGTM, last nitpick about the SPDXID

chmeliik:
> The linking to the outside is supposed to be represented by the `externalRefs` field, where PURLs are specified, as represented [here](https://github.com/RedHatProductSecurity/security-data-guidelines/blob/main/sbom/examples/container_image/build/ubi9-micro-container-9.4-6.1716471860.spdx.json).

That's what I was thinking, makes sense to me. However, I have a feeling that the arch-specific SBOMs do not include the necessary PURL anywhere. This is not something to be solved within this PR, but may need to be solved within the scope of the epic

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
BorekZnovustvoritel: The second point is an interesting idea. However I think it might slightly decrease the human-readability of the SBOM as the relationships will be less clear at a first sight. If you don't protest against this, I think I will just make the `Image` lowercase.

chmeliik: works for me :+1:

chmeliik (approved):
LGTM

Would you mind squashing the commits? We generally prefer commits to be atomic

BorekZnovustvoritel:
Understandable, squashed into a single commit. Thank you for your assistance

PR #177, opened 2024-11-05
Migrate to using Trusted Artifacts

mkosiarc:
Otherwise, we wont pass the release EC checks.
I just followed
https://konflux-ci.dev/docs/advanced-how-tos/using-trusted-artifacts/

and

https://github.com/konflux-ci/build-ta-pipeline-migration


.tekton/source-container-build-pull-request.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
     pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch == "main" && ( "source-container-build/***".pathChanged() || ".tekton/source-container-build-pull-request.yaml".pathChanged() )
-  creationTimestamp: null
+  creationTimestamp:
```
mkosiarc, line 11: Not sure if this is intended or just a bug in the migration script

mkosiarc: but the pipeline is passing, so I will leave it be I guess

mmorhun (approved):

PR #181, opened 2024-11-13
Add a script for sbom enrichment

Allda:
The script is responsible for expanding a image sbom by adding a reference pointing to the image into a SBOM content.

The script recognize an input format and adds necessary data into the list of components or packages. The script also sets a name of the SBOM based on a pullspec.

JIRA: ISV-5411, ISV-5320

### DIFF
--- a/sbom-utility-scripts/scripts/add-image-reference-script/README.md
+++ b/sbom-utility-scripts/scripts/add-image-reference-script/README.md
@@ -1,0 +1,144 @@
+# Add image reference script
+
+The script aims to enrich the SBOM file with additional information about the output image used in the build process.
+Based on a input SBOM type, the script updates certain fields with the image reference information. This is needed
+to provide a complete SBOM file that can be used for further analysis.
+
+## Usage
+
+```bash
+python add_image_reference.py \
+    --input-file ./input-sbom.json \
+    --output-file ./updated-sbom.json \
+    --image-url quay.io/foo/bar/ubi8:1.1 \
+    --image-digest sha256:011ff0cd8f34588d3eca86da97619f7baf99c8cc12e24cc3a7f337873c8d36cc
+```
+The script stores the updated SBOM in the output path provided.
+
+## List of updates
+### SPDX
+The script updates the following fields in the SPDX SBOM:SHA256
+- `packages` - the script adds a new package with the image reference information
+- `relationships` - the script adds a new relationship between the package and the image reference
+- `name` - the script adds the image reference as a name
+
+The logic of adding new packages and updating replationships is visualized in the following diagram:
+```
+example SPDX SBOM:
+
+                ROOT
+             /        \
+            /          \
+        DESCRIBES   DESCRIBES
+          /              \
+         /                \
+<pip main package>  <npm main package>
+        |                  |
+        |                  |
+     CONTAINS           CONTAINS
+        |                  |
+        |                  |
+    <pip deps>         <npm deps>
+
+
+SBOM after enrichment:
+
+                ROOT
+                 |
+                 |
+              DESCRIBES
+                 |
+                 |
+             <container>
+             /        \
+            /          \
+        CONTAINS    CONTAINS
+          /              \
+         /                \
+<pip main package>  <npm main package>
+        |                  |
+        |                  |
+     CONTAINS           CONTAINS
+        |                  |
+        |                  |
+    <pip deps>         <npm deps>
+
+```
+
+#### Example
+```json
+{
+  "name": "quay.io/foo/bar/ubi8@sha256:011ff0cd8f34588d3eca86da97619f7baf99c8cc12e24cc3a7f337873c8d36cc",
+  "packages": [
+    {
+      "SPDXID": "SPDXRef-image",
+      "name": "ubi8",
+      "versionInfo": "1.1",
+      "downloadLocation": "NOASSERTION",
+      "licenseConcluded": "NOASSERTION",
+      "supplier": "NOASSERTION",
+      "externalRefs": [
+        {
+          "referenceLocator": "pkg:oci/ubi8@sha256:011ff0cd8f34588d3eca86da97619f7baf99c8cc12e24cc3a7f337873c8d36cc?repository_url=quay.io/foo/bar/ubi8",
+          "referenceType": "purl",
+          "referenceCategory": "PACKAGE-MANAGER"
+        }
+      ],
+      "checksums": [
+        {
+          "algorithm": "SHA256",
+          "checksumValue": "011ff0cd8f34588d3eca86da97619f7baf99c8cc12e24cc3a7f337873c8d36cc"
+        }
+      ]
+    },
+  ],
+  "relationships": [
+    {
+      "spdxElementId": "SPDXRef-DOCUMENT",
+      "relationshipType": "DESCRIBES",
+      "relatedSpdxElement": "SPDXRef-image"
+    },
+  ]
+}
+
+```
+
+### CycloneDX
+- `components` - the script adds a new component with the image reference information
+- `metadata.component` - the script adds the image reference as a metadata.component
+
+#### Example
+```json
+{
+  "metadata": {
+    "component": {
+      "type": "container",
+      "name": "ubi8",
+      "purl": "pkg:oci/ubi8@sha256:011ff0cd8f34588d3eca86da97619f7baf99c8cc12e24cc3a7f337873c8d36cc?repository_url=quay.io/foo/bar/ubi8",
+      "version": "1.1",
+      "publisher": "Red Hat, Inc.",
+      "hashes": [
+        {
+          "alg": "SHA-256",
+          "content": "011ff0cd8f34588d3eca86da97619f7baf99c8cc12e24cc3a7f337873c8d36cc"
+        }
+      ]
+    },
+  },
+  "components": [
+    {
+      "type": "container",
+      "name": "ubi8",
+      "purl": "pkg:oci/ubi8@sha256:011ff0cd8f34588d3eca86da97619f7baf99c8cc12e24cc3a7f337873c8d36cc?repository_url=quay.io/foo/bar/ubi8",
+      "version": "1.1",
+      "publisher": "Red Hat, Inc.",
+      "hashes": [
+        {
+          "alg": "SHA-256",
+          "content": "011ff0cd8f34588d3eca86da97619f7baf99c8cc12e24cc3a7f337873c8d36cc"
+        }
+      ]
+    },
+  ]
+}
+```
--- a/sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
+++ b/sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
@@ -1,0 +1,404 @@
+#!/usr/bin/env python3
+import argparse
+import json
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Optional
+
+from packageurl import PackageURL
+
+
+@dataclass
+class Image:
+    repository: str
+    name: str
+    digest: str
+    tag: str
+
+    @staticmethod
+    def from_image_index_url_and_digest(
+        image_url_and_tag: str,
+        image_digest: str,
+    ) -> "Image":
+        """
+        Create an instance of the Image class from the image URL and digest.
+
+        Args:
+            image_url_and_tag (str): Image URL in the format 'registry.com/repository/image:tag'.
+            image_digest (str): Manifest digest of the image. (sha256:digest)
+
+        Returns:
+            Image: An instance of the Image class representing the image.
+        """
+        repository, tag = image_url_and_tag.rsplit(":", 1)
+        _, name = repository.rsplit("/", 1)
+        return Image(
+            repository=repository,
+            name=name,
+            digest=image_digest,
+            tag=tag,
+        )
+
+    @property
+    def digest_algo_cyclonedx(self) -> str:
+        """
+        Get the digest algorithm used for the image in cyclonedx format.
+        The output is in uppercase.
+
+        Returns:
+            str: Algorithm used for the digest.
+        """
+        algo, _ = self.digest.split(":")
+        mapping = {"sha256": "SHA-256", "sha512": "SHA-512"}
+        return mapping.get(algo, algo.upper())
+
+    @property
+    def digest_algo_spdx(self) -> str:
+        """
+        Get the digest algorithm used for the image in SPDX format.
+
+        Returns:
+            str: Algorithm used for the digest in SPDX format.
+        """
+        algo, _ = self.digest.split(":")
+        return algo.upper()
+
+    @property
+    def digest_hex_val(self) -> str:
+        """
+        Get the digest value of the image in hexadecimal format.
+
+        Returns:
+            str: Digest value in hexadecimal format.
+        """
+        _, val = self.digest.split(":")
+        return val
+
+    def purl(self) -> str:
+        """
+        Get the Package URL (PURL) for the image.
+
+        Returns:
+            str: A string representing the PURL for the image.
+        """
+        return PackageURL(
+            type="oci",
+            name=self.name,
+            version=self.digest,
+            qualifiers={"repository_url": self.repository},
+        ).to_string()
+
+
+def setup_arg_parser() -> argparse.ArgumentParser:
+    """
+    Setup the argument parser for the script.
+
+    Returns:
+        argparse.ArgumentParser: Argument parser for the script.
+    """
+    parser = argparse.ArgumentParser(description="Add image reference to image SBOM.")
+    parser.add_argument(
+        "--image-url",
+        type=str,
+        help="Image URL in the format 'registry.com/repository/image:tag'.",
+        required=True,
+    )
+    parser.add_argument(
+        "--image-digest",
+        type=str,
+        help="Image manifest digest in a form sha256:xxxx.",
+        required=True,
+    )
+    parser.add_argument(
+        "--input-file",
+        "-i",
+        type=Path,
+        help="SBOM file in JSON format.",
+        required=True,
+    )
+    parser.add_argument(
+        "--output-file",
+        "-o",
+        type=str,
+        help="Path to save the output SBOM in JSON format.",
+    )
+    return parser
+
+
+def update_component_in_cyclonedx_sbom(sbom: dict, image: Image) -> dict:
+    """
+    Update the CycloneDX SBOM with the image reference.
+
+    The reference to the image is added to the SBOM in the form of a component and
+    purl is added to the metadata.
+
+    Args:
+        sbom (dict): SBOM in JSON format.
+        image (Image): An instance of the Image class that represents the image.
+
+    Returns:
+        dict: Updated SBOM with the image reference added.
+    """
+    # Add the image component to the components list
+    image_component = {
+        "type": "container",
+        "name": image.name,
+        "purl": image.purl(),
+        "version": image.tag,
+        "hashes": [{"alg": image.digest_algo_cyclonedx, "content": image.digest_hex_val}],
+    }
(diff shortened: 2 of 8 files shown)


Allda:
@chmeliik Hey, this is the prerequisite of ISV-5411. Once this is merged I'll continue on the buildah task and remove embedded sbom in image as we agreed.
Please assign any reviewers to this PR based on your review procedure. Thanks!!

sbom-utility-scripts/scripts/add-image-reference-script/README.md
```diff
@@ -0,0 +1,95 @@
...
+    },
+  ]
+}
+
+```
+
+### CycloneDX
+- `components` - the script adds a new component with the image reference information
+- `metadata.component.purl` - the script adds the image reference as a purl
+- `metadata.component.name` - the script adds the image reference as a name
```
chmeliik, line 67: IMO we don't need to add the component to `components`, it should be enough that it's in `metadata.component`

chmeliik: s/don't need to/shouldn't/

My interpretation is that `components` is for dependencies, so the main component shouldn't be there

sbom-utility-scripts/scripts/add-image-reference-script/README.md
```diff
@@ -0,0 +1,95 @@
+# Add image reference script
+
+The script aims to enrich the SBOM file with additional information about the images used in the build process.
```
chmeliik, line 3: > information about the images used in the build process

To me this gives the false impression that this script handles the `FROM` images in the Containerfile (there's a different script for that in this repo). "Information about the output image" might be more accurate

sbom-utility-scripts/scripts/add-image-reference-script/README.md
```diff
@@ -0,0 +1,95 @@
...
+Based on a input SBOM type, the script updates certain fields with the image reference information. This is needed
+to provide a complete SBOM file that can be used for further analysis.
+
+## Usage
+
+```bash
+python add_image_reference.py \
+    --input-file ./input-sbom.json \
+    --output-path ./updated-sbom.json \
```
chmeliik, line 12: nitpick: let's use `*-file` or `*-path` consistently

sbom-utility-scripts/scripts/add-image-reference-script/README.md
```diff
@@ -0,0 +1,95 @@
...
+      "purl": "pkg:oci/ubi8@sha256:011ff0cd8f34588d3eca86da97619f7baf99c8cc12e24cc3a7f337873c8d36cc?arch=amd64&repository_url=quay.io/foo/bar/ubi8",
+      "version": "1.1",
+      "publisher": "Red Hat, Inc.",
+      "hashes": [
+        {
+          "alg": "SHA-256",
+          "content": "011ff0cd8f34588d3eca86da97619f7baf99c8cc12e24cc3a7f337873c8d36cc"
+        }
+      ]
+    },
```
chmeliik, line 92: Related to previous comment: this whole thing can be in `.metadata.component` instead (the attribute has the same schema as the elements in `.components[]`

sbom-utility-scripts/scripts/add-image-reference-script/README.md
Allda: This was requested by the TPA team since a missing reference under `components`  causes that SBOM in CycloneDX format is not analyzed correctly by the Guac and the SBOM is then ignored by the incident management process.

chmeliik: That sounds like a bug in Guac :thinking:

sbom-utility-scripts/scripts/add-image-reference-script/README.md
Allda: Fixed

sbom-utility-scripts/scripts/add-image-reference-script/README.md
Allda: Fixed

sbom-utility-scripts/scripts/add-image-reference-script/README.md
Allda: There are more use cases for having the image reference in a `components` field that can't be achieved by having the info in `metadata.component`. One of the limitations of the suggested field is that it can't have multiple purls.

Feel free to continue with this discussion here https://redhat-internal.slack.com/archives/C07MHR3MGLR/p1729693977566349.

For incident management and the CAMP project that is targeting the end of Q4, having the item in `components` seems to be a must-have requirement. The TPA is going to release a new 2.x version early next year, where this might be fixed. In the meantime, let's enable this and evaluate it later if you think it is an important detail.

chmeliik: Ack, sounds reasonable

chmeliik (commented):
Mostly LGTM, some minor concerns

midnightercz:
I don't think approach for spdx is correct. Every spdx document on the input (produced either by cachi2 or syft) will have "virtual" or "wrapper "package (I don't have better name for it) in relationship ROOT DESCRIBES SPDXRef-VirtualPackage and then other packages representing real packages are in relation SPDXRef-VirtualPackage CONTAINS SPDXRef-Package-A. Build dependencies (builder images) are in relation SPDXRef-Builder-Image1 BUILD_TOOL_OF SPDXRef-VirtualPackage.
So this virtual package represent the source which was used to created the sbom (container, directory, file,...). You can see here: https://github.com/konflux-ci/architecture/pull/213/files#diff-71b891db9bf4d748c67ae1b5779e6d57ad3fc384f672fae94b3b2627d8770679R149 how syft names it.
And if I undestand this PR correctly, you basically want to change the SBOM document to express that source of the sbom document is not some VirtualPackage (or in case of syft generated sboms SPDXRef-DocumentRoot-Directory-<dir>-) but it's specific container with name and digest and purl and other attributes.
If that's correct assumption I think you need to replace existing "VirtualPackage" with SPDXRef-Image instead of adding new package to the list of packages.
Btw I tried to run syft convert on cdx sbom with metadata.component to see how syft converts it so we can maybe get some hints there, but it doesn't convert it at all

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
```diff
@@ -0,0 +1,279 @@
...
+
+    Args:
+        sbom (dict): SBOM in JSON format.
+        image (Image): An instance of the Image class that represents the image.
+
+    Returns:
+        dict: Updated SBOM with the image reference added.
+    """
+    # First, update the metadata with the image purl
+    sbom["metadata"]["component"]["purl"] = image.purl()
```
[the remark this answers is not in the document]
Allda: Based on a quick check of the schema it should work. I'll try it.

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
```diff
@@ -0,0 +1,279 @@
...
+
+    # Then, add the image component to the components list
+    image_component = {
+        "type": "container",
+        "bom-ref": image.purl(),
+        "name": image.name,
+        "purl": image.purl(),
+        "version": image.tag,
+        "publisher": "Red Hat, Inc.",
```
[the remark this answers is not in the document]
Allda: Sure, no problem with that.

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
```diff
@@ -0,0 +1,279 @@
...
+    Returns:
+        dict: Updated SBOM with the image reference added.
+    """
+    # First, update the metadata with the image purl
+    sbom["metadata"]["component"]["purl"] = image.purl()
+
+    # Then, add the image component to the components list
+    image_component = {
+        "type": "container",
+        "bom-ref": image.purl(),
```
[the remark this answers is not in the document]
Allda: Fixed

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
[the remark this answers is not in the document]
Allda: Fixed

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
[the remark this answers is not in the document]
Allda: Fixed

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
```diff
@@ -0,0 +1,279 @@
...
+
+    Args:
+        sbom (dict): SBOM in JSON format.
+        image (Image): An instance of the Image class that represents the image.
+
+    Returns:
+        dict: Updated SBOM with the name field updated.
+    """
+    if sbom.get("bomFormat") == "CycloneDX":
+        sbom["metadata"]["component"]["name"] = f"{image.repository}@{image.digest}"
```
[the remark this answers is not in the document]
Allda: Removed

chmeliik:
Hypothetically, I can imagine a situation where an SPDX SBOM `DESCRIBES` two or more packages.

Using `cachi2` as an example (this is probably not what the SBOMs will actually look like, just for illustration):

```
input: [{"type": "pip"}, {"type": "npm"}]


cachi2 SPDX SBOM:

                ROOT
             /        \
            /          \
        DESCRIBES   DESCRIBES
          /              \
         /                \
<pip main package>  <npm main package>
        |                  |
        |                  |
     CONTAINS           CONTAINS
        |                  |
        |                  |
    <pip deps>         <npm deps>


SBOM after enrichment?

                ROOT
                 |
                 |
              DESCRIBES
                 |
                 |
             <container>
             /        \
            /          \
        CONTAINS    CONTAINS
          /              \
         /                \
<pip main package>  <npm main package>
        |                  |
        |                  |
     CONTAINS           CONTAINS
        |                  |
        |                  |
    <pip deps>         <npm deps>

```

chmeliik:
The most general handling could look something like this:

* for all `ROOT DESCRIBES some_pkg`
  * if `some_pkg` doesn't look like a "virtual" element (e.g. has a name and/or version),
    * replace `ROOT DESCRIBES some_pkg` with `container CONTAINS some_pkg`
  * else
    * remove `ROOT DESCRIBES some_pkg`
    * for all remaining relations of `some_pkg`, replace `some_pkg` with `container`
* add `ROOT DESCRIBES container`

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
```diff
@@ -0,0 +1,276 @@
...
+        Get the Package URL (PURL) for the image.
+
+        Returns:
+            str: A string representing the PURL for the image.
+        """
+        return PackageURL(
+            type="oci",
+            name=self.name,
+            version=self.digest,
+            qualifiers={"arch": self.arch, "repository_url": self.repository},
```
chmeliik, line 92: Cross-referencing this with the index_image_sbom script, I think we should drop the `arch` qualifier

The index_image SBOM uses purls like this

```
pkg:oci/name@sha256:{index_digest}?arch=amd64
pkg:oci/name@sha256:{amd64_manifest_digest}

pkg:oci/name@sha256:{index_digest}?arch=s390x
pkg:oci/name@sha256:{s390x_manifest_digest}
```

What we're creating here is
```
pkg:oci/name@sha256:{amd64_manifest_digest}?arch=amd64
```

Which wouldn't match any purl in the index image SBOM

sbom-utility-scripts/scripts/add-image-reference-script/README.md
```diff
@@ -0,0 +1,95 @@
...
+- `metadata.component.name` - the script adds the image reference as a name
+
+#### Example
+```json
+{
+  "metadata": {
+    "component": {
+      "name": "quay.io/foo/bar/ubi8@sha256:011ff0cd8f34588d3eca86da97619f7baf99c8cc12e24cc3a7f337873c8d36cc",
+      "purl": "pkg:oci/ubi8@sha256:011ff0cd8f34588d3eca86da97619f7baf99c8cc12e24cc3a7f337873c8d36cc?arch=amd64&repository_url=quay.io/foo/bar/ubi8"
+    }
```
chmeliik, line 76: Needs an update to match the new code

sbom-utility-scripts/scripts/add-image-reference-script/README.md
```diff
@@ -0,0 +1,95 @@
...
+      "externalRefs": [
+        {
+          "referenceLocator": "pkg:oci/ubi8@sha256:011ff0cd8f34588d3eca86da97619f7baf99c8cc12e24cc3a7f337873c8d36cc?repository_url=quay.io/foo/bar/ubi8",
+          "referenceType": "purl",
+          "referenceCategory": "PACKAGE-MANAGER"
+        }
+      ],
+      "checksums": [
+        {
+          "algorithm": "SHA256",
```
chmeliik, line 47:
```suggestion
          "algorithm": "SHA-256",
```

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
Allda: That makes sense. Since we don't have an index digest at this point in the pipeline I'll keep the arch-specific manifest and remove the arch qualifier.

sbom-utility-scripts/scripts/add-image-reference-script/README.md
Allda: The allowed value based on the schema is `SHA256` without a `-`. https://spdx.github.io/spdx-spec/v2.3/package-information/#710-package-checksum-field

sbom-utility-scripts/scripts/add-image-reference-script/README.md
Allda: Fixed

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
Allda: Fixed

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
```diff
@@ -0,0 +1,404 @@
...
+        virtual_root (str): A virtual root node identifier that needs to be replaced.
+        new_root (str): A new root node identifier that will replace the virtual root node.
+
+    Returns:
+        dict: Updated SBOM with the virtual root node replaced.
+    """
+    for relationship in sbom["relationships"]:
+        if relationship["spdxElementId"] == virtual_root:
+            relationship["spdxElementId"] = new_root
+
```
chmeliik, line 208: Should we do this for the `relatedSpdxElement` as well? The virtual root can also be on the "right side" of relationships, e.g. `some_package BUILD-TOOL-OF virtual_root`

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
```diff
@@ -0,0 +1,404 @@
...
+    """
+    Check if the package is a virtual - usually a package with empty values.
+
+    Args:
+        package (dict): A package element from the SBOM.
+
+    Returns:
+        bool: A boolean indicating if the package is a virtual root.
+    """
+    return package.get("name") == ""
```
chmeliik, line 254:
```suggestion
    return not package.get("name")
```
(for missing `name`, the current code would return False)

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
```diff
@@ -0,0 +1,404 @@
...
+
+    Args:
+        sbom (dict): SBOM in JSON format.
+        image (Image): An instance of the Image class that represents the image.
+
+    Returns:
+        dict: Updated SBOM with the image reference added.
+    """
+    # First, update the metadata with the image purl
+    sbom["metadata"]["component"]["purl"] = image.purl()
```
chmeliik, line 143: nitpick: this line isn't needed anymore

sbom-utility-scripts/scripts/add-image-reference-script/README.md
```diff
@@ -0,0 +1,146 @@
...
+- `components` - the script adds a new component with the image reference information
+- `metadata.component` - the script adds the image reference as a metadata.component
+
+#### Example
+```json
+{
+  "metadata": {
+    "component": {
+      "type": "container",
+      "bom-ref": "pkg:oci/ubi8@sha256:011ff0cd8f34588d3eca86da97619f7baf99c8cc12e24cc3a7f337873c8d36cc?repository_url=quay.io/foo/bar/ubi8",
```
chmeliik, line 116: nitpick: the README shouldn't include `bom-ref` anymore either

sbom-utility-scripts/scripts/add-image-reference-script/README.md
```diff
@@ -0,0 +1,146 @@
...
+## List of updates
+### SPDX
+The script updates the following fields in the SPDX SBOM:SHA256
+- `packages` - the script adds a new package with the image reference information
+- `relationships` - the script adds a new relationship between the package and the image reference
+- `name` - the script adds the image reference as a name
+
+The logic of adding new packages and updating replationships is visualized in the following diagram:
+```
+cachi2 SPDX SBOM:
```
chmeliik, line 27: nitpick: it's not clear if cachi2 SBOMs will look like this, let's just call it an "example SPDX SBOM"

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
Allda: Fixed

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
Allda: Removed

sbom-utility-scripts/scripts/add-image-reference-script/README.md
Allda: Removed

sbom-utility-scripts/scripts/add-image-reference-script/README.md
Allda: Renamed

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
Allda: I added a logic that covers an other direction.

sbom-utility-scripts/scripts/add-image-reference-script/tox.ini
```diff
@@ -0,0 +1,19 @@
+[tox]
```
chmeliik, line 1: Just noticed the GH workflow probably needs an update to run this, similar to https://github.com/konflux-ci/build-tasks-dockerfiles/blob/4770c7c4fedd305291936056392e877260968719/.github/workflows/build-sbom-utility-scripts-image.yml#L48-L52

chmeliik (commented):
One CI thing, otherwise LGTM

sbom-utility-scripts/scripts/add-image-reference-script/tox.ini
Allda: Added

chmeliik:
The CI is failing with
```
/home/runner/work/_temp/3426a099-da9e-4189-9ab2-8368e6cce995.sh: line 5: -: command not found
```
indentation problem I think

chmeliik (approved):
LGTM :+1:

mmorhun (approved):

chmeliik:
The pending Konflux check is a false positive, this part of the code has no Konflux pipeline. Merging

PR #191, opened 2024-11-28
Improve generation of base images SBOMs

mkosiarc:
The process how the base image SBOM is generated was slightly changed.

The most important change is due to requiring different inputs. The script now requires the full json of parsed Dockerfile, so we can extract the base images better. This allows us to extract base images data even for stage aliases.

Another advantage is that we are counting
the stages/layers more carefully, because we have more information about which stage was from scratch or from oci-archive.

KFLUXBUGS-1718


sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -87,23 +105,77 @@ def get_base_images_sbom_components(base_images_digests, is_last_from_scratch):
...
+    # info about base images
+    stages = parsed_dockerfile["Stages"]
+
+    for stage in stages:
+        if "Image" in stage["From"]:
+            base_images.append(stage["From"]["Image"])
+        elif "Scratch" in stage["From"]:
+            base_images.append("scratch")
+        elif "Stage" in stage["From"]:
+            stage_refering_to_index = stage["From"]["Stage"]["Index"]
```
chmeliik, line 150: nitpick: I thing `stage_index` would be clearer. `stage_refering_to_index` sounds like the referenced Stage object, not the index

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -87,23 +105,77 @@ def get_base_images_sbom_components(base_images_digests, is_last_from_scratch):
...
+            # Find the original stage/image. Named stage can refer to another named stage,
+            # so continue looking until we find the image those stages refer to.
+            while stage_refering_to_index is not None and stage_refering_to_index != {}:
+                refered_stage = stages[stage_refering_to_index]
+                stage_refering_to_index = refered_stage.get("From").get("Stage", {}).get("Index", None)
+                if stage_refering_to_index is None or stage_refering_to_index == {}:
+                    base_images.append(refered_stage["From"]["Image"])
```
chmeliik, line 157: Is it technically possible to have `FROM scratch AS somestage`? When we find the original stage, we should probably apply the same image-getting logic to it that we apply to every stage

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -116,21 +188,28 @@ def main():
...
+    base_images = get_base_images_from_dockerfile(parsed_dockerfile)
 
-    is_last_from_scratch = False
-    if base_images_from_dockerfile[-1] == "scratch":
-        is_last_from_scratch = True
+    base_images_digests_raw = args.base_images_digests.read_text().splitlines()
+    base_images_digests = {}
+    for item in base_images_digests_raw:
+        image_reference, digest = item.split("|")
+        base_images_digests[image_reference] = digest
```
tkdchen, line 200: Can we shorten these lines as following?

```python
base_images_digests = dict(item.split("|") for item in base_images_digests_raw)
```

mkosiarc: updated

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -33,29 +33,47 @@ def parse_image_reference_to_parts(image):
...
 
 
-def get_base_images_sbom_components(base_images_digests, is_last_from_scratch):
+def get_base_images_sbom_components(base_images, base_images_digests):
     """
     Creates the base images sbom data
 
-    :param base_images_digests: (List) - list of base images digests, same as BASE_IMAGE_DIGESTS tekton result
-    :param is_last_from_scratch: (Boolean) - Is the last stage/base image from scratch?
+    :param base_images: (List) - List of base images used during build, in the order they were used.
```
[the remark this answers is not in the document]
mkosiarc: updated

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -33,29 +33,47 @@ def parse_image_reference_to_parts(image):
...
+    :param base_images_digests: (Dict) - Dict of base images digests, where the key is the image reference as
+                                         used in the original Dockerfile and the values are the full image reference
+                                         with digests that was actually used by buildah during build time.
```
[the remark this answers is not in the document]
mkosiarc: I updated the text slightly and added the examples, so it should be more clear now.

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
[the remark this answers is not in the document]
mkosiarc: added

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
mkosiarc: I really wanted to convey that is is not the stage of the current index, but that it is a reference to another stage. But changed it to `stage_index` and the rest is clearer from context.

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
mkosiarc: Hmm, that perversion is indeed possible. But that brings another issue, for example
```
FROM scratch AS stageA
COPY . .

FROM stageA
COPY . .
```
When parsed, it does not behave well - e.g. the stageA, does not hold a index reference to the original from scratch, but only has "Image", which is not what that is at all, so it is not even parsed "properly" and it would be hard for us to start deciding whether is is just a perversion - reference to FROM stage or an actual image. That would bring another layer of complexity unnecessarily I would say. Nobody should ever use that :D  
```
{
    "Stages": [
        {
            "BaseName": "scratch",
            "As": "stageA",
            "From": {"Scratch": True},
        },
        {
            "BaseName": "stageA",
            "From": {"Image": "stageA"},
        },
    ]
},
```

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -33,29 +33,47 @@ def parse_image_reference_to_parts(image):
...
 
-    # property_name shows whether the image was used only in the building process
-    # or if it is the final base image. If the final base image is scratch, then
-    # this is omitted, because we aren't including scratch in the sbom.
-    for index, image in enumerate(base_images_digests):
+    for index, image in enumerate(base_images):
+        # flatpak archive and scratch are not real base images. So we skip them, but
+        # in a way that allows us to keep the correct track of index variable that
+        # refers to stage number.
+        if image.startswith("oci-archive") or image == "scratch":
```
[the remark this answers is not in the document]
mkosiarc: if we have only from
```
FROM oci-archive
```
It assumes that it is a real image, and starts pulling from registries configured as default. So in theory, oci-archive would be a valid image and ideally, we would want to make our parsing more complexes and differentiate between a real image called oci-archive and a flatpak. But I would say that's not worth getting into, especially in the view of https://github.com/containers/buildah/issues/5529, that suggests that flatpaks might be handled differently in the future.

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
chmeliik: Ack, there probably shouldn't be any legitimate use case for this

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
```diff
@@ -87,23 +114,77 @@ def get_base_images_sbom_components(base_images_digests, is_last_from_scratch):
...
+        elif "Scratch" in stage["From"]:
+            base_images.append("scratch")
+        elif "Stage" in stage["From"]:
+            stage_index = stage["From"]["Stage"]["Index"]
+            # Find the original stage/image. Named stage can refer to another named stage,
+            # so continue looking until we find the image those stages refer to.
+            while stage_index is not None and stage_index != {}:
+                refered_stage = stages[stage_index]
+                stage_index = refered_stage.get("From").get("Stage", {}).get("Index", None)
+                if stage_index is None or stage_index == {}:
```
chmeliik, line 165: nitpick: I don't think `stage_index` will ever be a dict
```suggestion
            while stage_index is not None:
                refered_stage = stages[stage_index]
                stage_index = refered_stage.get("From").get("Stage", {}).get("Index", None)
                if stage_index is None:
```

mkosiarc:
switched to using empty space, instead of "|" as discussed in https://github.com/konflux-ci/build-definitions/pull/1659

chmeliik (approved):
LGTM with one nitpick

sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py
mkosiarc: you are right, the following `.get` handles that and it will be just None. Good catch, updated.

PR #192, opened 2024-11-29
chore: remove workflow which don't do anything useful

tisutisu:
This part of the code is a leftover, don't see doing anything useful, so removing


chmeliik (approved):
Thanks!

mkosiarc (approved):

PR #207, opened 2024-12-19
fix: change query to find source_build image

tisutisu:
with new component being added to konflux, snapshot contains more images, this is fix the query to find correct image


source-container-build/README.md
```diff
@@ -34,3 +34,5 @@ push event (when a pull request is merged).
 4. A new source build task bundle is built and pushed to registry.
 5. Components repositories receive bundle update pull request, which is sent
    during a Build service Renovate controller reconciliation.
+
+#temp change to trigger CI
```
mkosiarc, line 38: I think it would be a good idea if we triggered the pipeline if the e2e change, so we could validate it directly in the PR. Would you mind updating the cel-expressions for PR pipeline to trigger in these cases?

tisutisu: Good point. Updated the `on-cel-expression` to trigger it for test change.

mmorhun (approved):

PR #209, opened 2024-12-20
Add new image to inject image content manifest

ralphbean:
For backwards compatibility with scanners that still expect to find this file.

Related: https://github.com/konflux-ci/build-definitions/pull/1771

### DIFF
--- a/icm-injection-scripts/scripts/inject-icm.sh
+++ b/icm-injection-scripts/scripts/inject-icm.sh
@@ -14,8 +14,7 @@
 
 set -euo pipefail
 
-CONTAINER="${1}"
-IMAGE="${2}"
+IMAGE="${1}"
 SQUASH="${SQUASH:-false}"
 
 icm_filename="content-sets.json"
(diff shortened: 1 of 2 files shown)


icm-injection-scripts/scripts/inject-icm.sh
```diff
@@ -0,0 +1,68 @@
...
+# deprecated.
+#
+# This is only possible for images built hermetically with prefetch
+
+set -euo pipefail
+
+IMAGE="${1}"
+SQUASH="${SQUASH:-false}"
+
+icm_filename="content-sets.json"
```
MartinBasti, line 20: OSBS generates `<name>-<version>-<release>.json` name to avoid overwritting ICMs of parent images. Are we ok with overwriting parent images ICMs here?

ralphbean: Yeah, I asked Clair devs and they said it's ok. They're always looking at things layer by layer anyways - they only consult the ICM that happens to appear in the same layer they're considering.

MartinBasti (approved):

PR #214, opened 2025-01-03
fix(KONFLUX-6229) fix ICM injection for multiarch images

ralphbean:
This ICM injection step accepts the IMAGE param as the image to operate on, which needs to account for whether or not the IMAGE_APPEND_PLATFORM param is also set to true.

https://issues.redhat.com/browse/KONFLUX-6229

### DIFF
--- a/icm-injection-scripts/scripts/inject-icm.sh
+++ b/icm-injection-scripts/scripts/inject-icm.sh
@@ -17,6 +17,12 @@ set -euo pipefail
 IMAGE="${1}"
 SQUASH="${SQUASH:-false}"
 
+set +u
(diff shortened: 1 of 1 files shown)


ralphbean:
This is a followup to https://github.com/konflux-ci/build-tasks-dockerfiles/pull/209.

ralphbean:
I think this is successfully validated https://github.com/konflux-ci/oras-container/pull/146 by way of https://github.com/konflux-ci/build-definitions/pull/1785

ralphbean:
Ah, but there is a better fix in https://github.com/konflux-ci/build-definitions/pull/1784. Let's use that!

PR #220, opened 2025-01-08
remove release from sbom-utility-scripts GH action

tnevrlka:
- Build and release of sbom-utility-scripts image is now done via Konflux
- The `Build Image` and `Push to Quay` steps are no longer necessary
- Tox checks are still useful as testing though
- Since the workflow is now not building anymore, rename to
  `test-sbom-utility-scripts`

### DIFF
--- a/.github/workflows/test-sbom-utility-scripts-image.yml
+++ b/.github/workflows/test-sbom-utility-scripts-image.yml
@@ -1,27 +1,22 @@
-name: Build sbom utility scripts image
+name: Test sbom utility scripts image
 
 on:
   push:
     branches:
       - main
     paths:
-      - .github/workflows/build-sbom-utility-scripts-image.yml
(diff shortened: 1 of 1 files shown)


.github/workflows/build-sbom-utility-scripts-image.yml
```diff
@@ -1,80 +0,0 @@
...
-      run: |
-        python3 -m pip install tox
-        cd ./sbom-utility-scripts/scripts/index-image-sbom-script/
-        tox
-
-    - name: Run tox checks for index-image-sbom-script
-      run: |
-        python3 -m pip install tox
-        cd ./sbom-utility-scripts/scripts/add-image-reference-script/
-        tox
```
chmeliik, line 58: Maybe we should keep this part? I don't think we have a Konflux replacement for running the unit tests

chmeliik (approved):

PR #222, opened 2025-01-08
merge_sboms: support SPDX

chmeliik:
Add SPDX support in the merge_sboms script. See the individual commits for more details.

This PR partially depends on https://github.com/containerbuildsystem/cachi2/pull/608 (mostly because that PR's version of cachi2 was used to generate the cachi2 SBOM in the test data). But it should be reviewable or even mergeable independently of the cachi2 PR


chmeliik:
As usual: 53776 of the added lines are test data. The changes you actually need to review are not very large.

chmeliik:
Added "handling" for the `.files` attribute in SPDX (more specifically, we just drop the whole `.files` when merging because we don't care about them)

mkosiarc (approved):

PR #223, opened 2025-01-09
inject-icm: support SPDX

chmeliik:
Detect the input SBOM format, use the right query to get the purls based on SBOM format

### DIFF
--- a/.github/workflows/inject-icm-test.yaml
+++ b/.github/workflows/inject-icm-test.yaml
@@ -1,0 +1,25 @@
+name: Test icm-injection-scripts
+
+on:
+  pull_request:
+    branches:
+    - main
+    paths:
+    - .github/workflows/inject-icm-test.yaml
+    - icm-injection-scripts/**
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Run test for cyclonedx
+      run: |
+        TEST_SBOM_FORMAT=cyclonedx ./icm-injection-scripts/scripts/test-inject-icm.sh
+
+    - name: Run test for spdx
+      run: |
+        TEST_SBOM_FORMAT=spdx ./icm-injection-scripts/scripts/test-inject-icm.sh
--- a/icm-injection-scripts/scripts/inject-icm.sh
+++ b/icm-injection-scripts/scripts/inject-icm.sh
@@ -31,6 +31,7 @@ base_image_digest=$(buildah inspect --format '{{ index .ImageAnnotations "org.op
 
 echo "Creating container from $IMAGE"
 CONTAINER=$(buildah from --pull-never $IMAGE)
+trap 'buildah rm "$CONTAINER"' EXIT
 
 echo "Preparing construction of $location for container $CONTAINER to be committed back as $IMAGE (squash: $SQUASH)"
 cat >content-sets.json <<EOF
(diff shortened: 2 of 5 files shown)


chmeliik:
How I tested this:

1. delete the part of the script that makes it hard to use locally
```diff
diff --git a/icm-injection-scripts/scripts/inject-icm.sh b/icm-injection-scripts/scripts/inject-icm.sh
index 622f40c..e7a17f1 100755
--- a/icm-injection-scripts/scripts/inject-icm.sh
+++ b/icm-injection-scripts/scripts/inject-icm.sh
@@ -25,14 +25,6 @@ if [ ! -f "./sbom-cachi2.json" ]; then
   exit 0
 fi
 
-echo "Extracting annotations to copy to the modified image"
-base_image_name=$(buildah inspect --format '{{ index .ImageAnnotations "org.opencontainers.image.base.name"}}' "$IMAGE" | cut -f1 -d'@')
-base_image_digest=$(buildah inspect --format '{{ index .ImageAnnotations "org.opencontainers.image.base.digest"}}' "$IMAGE")
-
-echo "Creating container from $IMAGE"
-CONTAINER=$(buildah from --pull-never $IMAGE)
-
-echo "Preparing construction of $location for container $CONTAINER to be committed back as $IMAGE (squash: $SQUASH)"
 cat >content-sets.json <<EOF
 {
     "metadata": {
```

2. clone https://github.com/containerbuildsystem/cachi2/pull/608
3. generate an rpms-included SBOM in SPDX format
```bash
cachi2 merge-sboms --sbom-output-type spdx tests/integration/test_data/rpm_e2e_test/bom.json tests/integration/test_data/rpm_e2e_test_module/bom.json > ~/RedHat/Stonesoup/build-tasks-dockerfiles/icm-injection-scripts/scripts/sbom-cachi2.json
```
4. run the script
```bash
$ bash inject-icm.sh foo
Constructed the following:
{
  "metadata": {
    "icm_version": 1,
    "icm_spec": "https://raw.githubusercontent.com/containerbuildsystem/atomic-reactor/master/atomic_reactor/schemas/content_manifest.json",
    "image_layer_index": 0
  },
  "from_dnf_hint": true,
  "content_sets": [
    "releases",
    "rpmfusion-free",
    "ubi-8-appstream-rpms",
    "ubi-8-appstream-source"
  ]
}
Writing that to /root/buildinfo/content_manifests/content-sets.json
inject-icm.sh: line 60: CONTAINER: unbound variable
```
5. Repeat steps 3 and 4 but use `--sbom-output-type cyclonedx` when generating the sbom-cachi2.json. As expected, the output was the same for SPDX and CycloneDX

chmeliik:
Rebased on main

chmeliik:
Added automated test for inject-icm.sh

tkdchen (commented):
LGTM.

A question is, what does it mean if `referenceLocator` is not included in one of the packages?

This change keeps silent. As the script handles SBOM output from other tool, for clarify, do you think it is necessary to log info or warning messages for the packages that do not have `referenceLocator`. If this is not a problem for handling SBOM, it's ok to keep the current behavior.

chmeliik:
> LGTM.
> 
> A question is, what does it mean if `referenceLocator` is not included in one of the packages?
> 
> This change keeps silent. As the script handles SBOM output from other tool, for clarify, do you think it is necessary to log info or warning messages for the packages that do not have `referenceLocator`. If this is not a problem for handling SBOM, it's ok to keep the current behavior.

According to the [schema](https://github.com/spdx/spdx-spec/blob/ed8c9b520963b651fce3c86422b387e92e6d9a8f/schemas/spdx-schema.json#L336), each item in the `.externalRefs` array must have `.referenceLocator`, so I don't think we need to worry too much about that.

It is also fairly normal for SPDX packages to not include the `externalRefs` array or to not include any item with `referenceType == "purl"` in the array. We want to skip those, and IMO it's not necessary to log them

tkdchen (approved):

PR #226, opened 2025-01-14
Add script for generating SBOM from oci-copy.yaml

chmeliik:
This should be a saner way to generate SBOMs from oci-copy.yaml compared to what the task is [currently doing](https://github.com/konflux-ci/build-definitions/blob/9f209df8c12e8b321e63b77b5c91cb1523080a59/task/oci-copy/0.1/oci-copy.yaml#L276). Especially now that SPDX support is needed as well.


ralphbean (approved):
Nice work!

chmeliik:
Forgot to run the unit tests for the new script in CI. Added in latest commit

mmorhun (approved):

PR #238, opened 2025-02-01
Fix merge_sboms.py formatting

tnevrlka:
New version of black is now failing due to a formatting issue


chmeliik (approved):
Nice, they fixed the formatting in `black` :tada:

PR #243, opened 2025-02-12
icm-injection-scripts: Use /usr/share/buildinfo by default

cgwalters:
Today this code writes to /root/buildinfo as a made up location from way back in the days of OSBS.

However, having this be in /root (/var/roothome) conflicts with bootc (RHEL Image Mode) that wants /var to start out as empty as possible in the container image (it acts as a volume/machine local state - https://containers.github.io/bootc/filesystem.html#var )

We actually hit this with RHEL CoreOS and introduced /usr/share/buildinfo - see https://github.com/coreos/coreos-assembler/commit/19531a6597eb4fc27beefea230002545e9a7cb11

And the scanners were taught to handle this (ref https://github.com/quay/claircore/commit/7250a059c82f019859e7493c3e2e452155da6164 )

This changes the script to just write to the /usr/share location by default, as it should still be compatible with scanners.

xref https://issues.redhat.com/browse/KONFLUX-6844

### DIFF
--- a/icm-injection-scripts/scripts/inject-icm.sh
+++ b/icm-injection-scripts/scripts/inject-icm.sh
@@ -18,7 +18,9 @@ IMAGE="${1}"
 SQUASH="${SQUASH:-false}"
 
 icm_filename="content-sets.json"
-location="/root/buildinfo/content_manifests/${icm_filename}"
+# Note this used to be /root/buildinfo/content_manifests but is now /usr/share/buildinfo for compatibility
+# with bootc/ostree systems. Ref https://issues.redhat.com/browse/KONFLUX-6844
+location="/usr/share/buildinfo/${icm_filename}"
 
 if [ ! -f "./sbom-cachi2.json" ]; then
   echo "Could not find sbom-cachi2.json. No content_sets found for ICM"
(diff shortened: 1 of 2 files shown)


chmeliik (approved):
Thanks!

cgwalters:
Fallout in https://issues.redhat.com/browse/STONEBLD-3715

cgwalters:
xref https://gitlab.com/fedora/bootc/base-images/-/merge_requests/574

chmeliik:
@cgwalters for what it's worth, I believe that Clair does support the /usr/share/buildinfo path these days: https://github.com/quay/claircore/blob/11f50838da39f2140f6603649f3b906433f77e78/rhel/rhcc/detector.go#L27

So setting [`ICM_KEEP_COMPAT_LOCATION`](https://github.com/konflux-ci/build-definitions/blob/e96f4c79387f635a1a49cb3298541ab25b3ddada/task/buildah/0.9/buildah.yaml#L93-L96)=false should be fine now

cgwalters:
The problem isn't Clair it was Prisma Defender, see https://redhat.atlassian.net/browse/RHEL-106895

PR #254, opened 2025-03-10
Add build image support to add_image_reference

Zokormazo:
The intent of this PR is to extend the funcionality of `add_image_reference.py` script to be able to inject build image references when the script is called with `--build-image` flag.

This is useful for https://github.com/konflux-ci/build-definitions/pull/2044 to be able to inject the image that was used to run the script as part of the build images on the final sbom.

The script's default functionality continues being the same when is executed without `--build-image` .

Calling it with the new flag will change the functionality to:

1.  Add a `properties` block to the package injected on the sbom. The resultant package will be something like this:

```
            "components": [
                {
                    "type": "container",
                    "name": "quay.io/mkosiarc_rhtap/single-container-app",
                    "purl": "pkg:oci/single-container-app@sha256:8f99627e843e931846855c5d899901bf093f5093e613a92745696a26b5420941?repository_url=quay.io/mkosiarc_rhtap/single-container-app",
                    "properties": [
                        {
                            "name": "konflux:container:is_builder_image:for_stage",
                            "value": "0",
                        }
                    ]
                }
```

2. For cyclone format:
    -  the package will not overwride the `metadata[component]` block
    - the package will be appended to the `formulation` section instead of `components` section

3. For SPDX format:
    - the package will not be colocated on the root of the document
    - the package will be injected with a `BUILD_TOOL_OF` relationship with the root package.

These changes are following the same format `base-images-sbom-script` uses to inject the base images coming from the Containerfile.

The main reason to colocate the functionality here and not on `base-images-sbom-script` is that the `base-images-sbom-script` logic is dependant on the content of the Dockerfile to inject the images. The changes the script would need to add arbitrary new build images that are not present on the Dockerfile would be big. Meanwhile the changes to repurpose this script to handle the original case and the new case are much smaller. An independent new script would be an option too, but the overlap of the code between `add_image_reference` and the new script would be huge.

TODO
- Finish SPDX tests
- Update README.md
- Update docstrings of the modified functions

### DIFF
--- a/sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
+++ b/sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
@@ -1,11 +1,18 @@
 #!/usr/bin/env python3
 import argparse
+import datetime
+import hashlib
 import json
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Optional
 
 from packageurl import PackageURL
+
+BUILDER_IMAGE_PROPERTY = {
+    "name": "konflux:container:is_builder_image:additional_builder_image",
+    "value": "script-runner-image",
+}
 
 
 @dataclass
@@ -14,11 +21,13 @@ class Image:
     name: str
     digest: str
     tag: str
+    is_builder_image: bool
 
     @staticmethod
     def from_image_index_url_and_digest(
         image_url_and_tag: str,
         image_digest: str,
+        is_builder_image: bool,
     ) -> "Image":
         """
         Create an instance of the Image class from the image URL and digest.
@@ -26,18 +35,14 @@ class Image:
         Args:
             image_url_and_tag (str): Image URL in the format 'registry.com/repository/image:tag'.
             image_digest (str): Manifest digest of the image. (sha256:digest)
+            is_builder_image (bool): Whether the image reference is a builder image or not. (False)
 
         Returns:
             Image: An instance of the Image class representing the image.
         """
         repository, tag = image_url_and_tag.rsplit(":", 1)
         _, name = repository.rsplit("/", 1)
-        return Image(
-            repository=repository,
-            name=name,
-            digest=image_digest,
-            tag=tag,
-        )
+        return Image(repository=repository, name=name, digest=image_digest, tag=tag, is_builder_image=is_builder_image)
 
     @property
     def digest_algo_cyclonedx(self) -> str:
@@ -73,6 +78,20 @@ class Image:
         """
         _, val = self.digest.split(":")
         return val
+
+    @property
+    def spdx_id(self) -> str:
+        """
+        Get the SPDXId value for the image.
+
+        Returns:
+            str: SPDXId value to the image
+        """
+        return (
+            f"SPDXRef-image-{self.name}-{hashlib.sha256(self.purl().encode()).hexdigest()}"
+            if self.is_builder_image
+            else "SPDXRef-image"
+        )
 
     def purl(self) -> str:
         """
@@ -122,6 +141,14 @@ def setup_arg_parser() -> argparse.ArgumentParser:
         type=str,
         help="Path to save the output SBOM in JSON format.",
     )
+    parser.add_argument(
+        "--builder-image",
+        "-b",
+        action=argparse.BooleanOptionalAction,
+        default=False,
+        type=bool,
+        help="Add the referenced image as a builder image.",
+    )
     return parser
 
 
@@ -139,7 +166,6 @@ def update_component_in_cyclonedx_sbom(sbom: dict, image: Image) -> dict:
     Returns:
         dict: Updated SBOM with the image reference added.
     """
-    # Add the image component to the components list
     image_component = {
         "type": "container",
         "name": image.name,
@@ -147,8 +173,15 @@ def update_component_in_cyclonedx_sbom(sbom: dict, image: Image) -> dict:
         "version": image.tag,
         "hashes": [{"alg": image.digest_algo_cyclonedx, "content": image.digest_hex_val}],
     }
-    sbom["components"].insert(0, image_component)
-    sbom["metadata"]["component"] = image_component
+    if image.is_builder_image:
+        # Add builder image property to image_component
+        image_component["properties"] = [BUILDER_IMAGE_PROPERTY]
+        # Add the builder image component to formulation section
+        sbom.setdefault("formulation", []).append({"components": [image_component]})
+    else:
+        # Add the image component to the components list
+        sbom["components"].insert(0, image_component)
+        sbom["metadata"]["component"] = image_component
     return sbom
 
 
@@ -305,6 +338,36 @@ def redirect_current_roots_to_new_root(sbom: dict, new_root: str) -> dict:
     return sbom
 
 
+def find_spdx_root_package(sbom: dict) -> str:
+    """
+    Find the element that's in relationship <DOCUMENT> DESCRIBES <ELEMENT>.
+
+    If there isn't exactly one such element, raise an error.
+
+    Args:
+        sbom (dict): SBOM in JSON format.
+    Returns:
+        str: SPDXID reference of the spdx root element
+    """
+    doc_spxid = sbom["SPDXID"]
+    doc_describes = [
+        r["relatedSpdxElement"]
+        for r in sbom.get("relationships", [])
+        if r["spdxElementId"] == doc_spxid and r["relationshipType"] == "DESCRIBES"
+    ]
+    if len(doc_describes) != 1:
+        raise ValueError(
+            "Expected to find exactly one <DOCUMENT> DESCRIBES <ROOT> relationship. "
+            f"Found {len(doc_describes)} ROOTs: {doc_describes}"
+        )
+    return doc_describes[0]
+
+
+def _datetime_utc_now() -> datetime.datetime:
+    # a mockable datetime.datetime.now (just for tests):
+    return datetime.datetime.now(datetime.UTC)  # pragma: no cover
+
+
 def update_package_in_spdx_sbom(sbom: dict, image: Image) -> dict:
     """
     Update the SPDX SBOM with the image reference.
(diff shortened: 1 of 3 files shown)


Zokormazo:
@Allda @mkosiarc @arewm good morning o/

Could I get some feedback on the code here before I finish the tests for spdx and update the doc strings?

Thanks in advance

mmorhun (commented):
Could you please make PR description explaining the changes, what how and why was done?

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
```diff
@@ -139,16 +149,25 @@ def update_component_in_cyclonedx_sbom(sbom: dict, image: Image) -> dict:
...
         "purl": image.purl(),
         "version": image.tag,
         "hashes": [{"alg": image.digest_algo_cyclonedx, "content": image.digest_hex_val}],
     }
-    sbom["components"].insert(0, image_component)
-    sbom["metadata"]["component"] = image_component
+    if image.build_image:
+        # Add base image property to image_component
+        image_component["properties"]: [{
+            "name": "konflux:container:is_builder_image:for_stage",
```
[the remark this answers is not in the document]
brunoapimentel: I believe this is following the same property established [here](https://github.com/konflux-ci/build-tasks-dockerfiles/blob/main/sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py#L98).

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
```diff
@@ -139,16 +149,25 @@ def update_component_in_cyclonedx_sbom(sbom: dict, image: Image) -> dict:
...
+            "value": "1",
```
brunoapimentel, line 163: I think hardcoding this as 1 wouldn't be correct, since this can be added to an image that already has several `konflux:container:is_builder_image:for_stage`, and potentially have one that is already stage 1. Maybe we'll need to increment this accordingly?

brunoapimentel (commented):
I'm not a direct maintainer of this script, but from my understanding, the `add_image_reference` script is currently meant for adding the output image to the SBOM, so the proposed change is stretching its goal and recontextualizing it a bit.

I don't think this is a bad choice, though, since the alternative would be to add it to the `base_images_sbom_script`, which only works by parsing the Containerfile and getting the base images from it. Trying to incorporate these changes to `base_images_sbom_script` would result in much bigger changes, and I don't see how to do it smoothly.

So I think that your approach is probably the best one, but we'll need to change the README to recontextualize the `add_image_reference` script's goal and document the new behavior.

brunoapimentel:
I was also wondering if it could be confusing to a user when he inspects the SBOM and finds an additional builder image that is not part of the Containerfile. I imagine that the fact that it is coming from an image used in a specific pipeline task would not be immediately obvious.

That being said, I don't think that SPDX or CycloneDX have any standard means of dealing with "additional images that were used as part of the build pipeline", so probably adding it as a builder image makes the most sense. It'd be nice if we could figure out any additional info that can be added to the SBOM to make this property meaning's clearer.

The only thing that comes to mind is to add the name of the task instead of a integer as value for `konflux:container:is_builder_image:for_stage`, but I still don't think it solves the issue perfectly. In any case, feel free to dismiss this comment, I am mostly thinking about the user experience here.

Zokormazo:
> I was also wondering if it could be confusing to a user when he inspects the SBOM and finds an additional builder image that is not part of the Containerfile. I imagine that the fact that it is coming from an image used in a specific pipeline task would not be immediately obvious.
> 
> That being said, I don't think that SPDX or CycloneDX have any standard means of dealing with "additional images that were used as part of the build pipeline", so probably adding it as a builder image makes the most sense. It'd be nice if we could figure out any additional info that can be added to the SBOM to make this property meaning's clearer.
> 
> The only thing that comes to mind is to add the name of the task instead of a integer as value for `konflux:container:is_builder_image:for_stage`, but I still don't think it solves the issue perfectly. In any case, feel free to dismiss this comment, I am mostly thinking about the user experience here.

This is a great appreciation, and I wondered the same thing :)

This image is not really part of the multi stage dockerfile, so for_stage and an integer does not codify properly what this image was used for. I don't have enough knowledge around sbom formats to make a decision here either :)

- Can we use something different to `konflux:container:is_builder_image:for_stage` ? Is there a field that suits better for this case?
- Can `konflux:container:is_builder_image:for_stage`  hold an string? Could `run-script` be the value on that case?
- If it can not hold an string, could we use `-1` ? Would this represent he use case better?

I need some help to answer those questions and make a decision here :)

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
Zokormazo: Yeah. A hardcoded 1 is wrong. the discussion about the property will condition what we do here :)
https://github.com/konflux-ci/build-tasks-dockerfiles/pull/254#issuecomment-2717609451

brunoapimentel:
>     * Can we use something different to `konflux:container:is_builder_image:for_stage` ? Is there a field that suits better for this case?

I believe we can. These `konflux:` properties were defined by us, since there was no [official property](https://github.com/CycloneDX/cyclonedx-property-taxonomy) that fit our use case.
 
>     * Can `konflux:container:is_builder_image:for_stage`  hold an string? Could `run-script` be the value on that case?

Yes, the value field type is string (CycloneDx [docs](https://cyclonedx.org/docs/1.6/json/#components_items_properties_items_value)).

As for SPDX, things are a little different. We tend to simply covert the CycloneDX into annotations, but if you look at the `add_image_reference` script, it is not being done there, and I assume it's because the component is clearly declared by making use of SPDX's relationships. I think `BUILD_DEPENDENCY_OF` still makes the most sense in our case (as it is used [here](https://github.com/konflux-ci/build-tasks-dockerfiles/blob/main/sbom-utility-scripts/scripts/base-images-sbom-script/app/base_images_sbom_script.py#L274)), but the full list is [here](https://spdx.github.io/spdx-spec/v2.3/relationships-between-SPDX-elements/), maybe you can find something more fitting (or we can simply dump the CDX property).

Zokormazo:
Since they are defined by us, it would make more sense to not use `for_stage` then?

Something like `konflux:container:is_builder_image:run_script` or ``konflux:container:is_builder_image:additional_build_image` , value: `run-script` ? for SPDX, `OTHER` and a description on the field could make more sense?

brunoapimentel:
> Since they are defined by us, it would make more sense to not use `for_stage` then?
> 
> Something like `konflux:container:is_builder_image:run_script` or `konflux:container:is_builder_image:additional_build_image` , value: `run-script` ? for SPDX, `OTHER` and a description on the field could make more sense?

Naming things is always hard :sweat: 

I still like `konflux:container:is_builder_image:additional_build_image` , value: `run-script` a little better, and this choice will allow us to add even additional builder images types (as values) in case any other pops up in the future. I'd just rename it to additional_build**er**_image, for consistency.

Zokormazo:
> I still like konflux:container:is_builder_image:additional_build_image , value: run-script a little better, and this choice will allow us to add even additional builder images types (as values) in case any other pops up in the future. I'd just rename it to additional_builder_image, for consistency.

Haha, just finished implementation and tests without this. Will update after lunch

Zokormazo:
@brunoapimentel changed the annotation name/value. wdyt now? Ready to start updating docstrings and the README ?

brunoapimentel:
> @brunoapimentel changed the annotation name/value. wdyt now? Ready to start updating docstrings and the README ?

It's fine by me. Any more considerations, @mmorhun?

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
```diff
@@ -1,24 +1,31 @@
...
+    "value": "run-script",
+}
 
 @dataclass
 class Image:
     repository: str
     name: str
     digest: str
     tag: str
+    builder_image: bool
```
mmorhun, line 22: nitpick: `is_builder_image` is more clear.

mmorhun (approved):

Zokormazo:
Per https://github.com/konflux-ci/build-definitions/pull/2044#discussion_r1995228161 , I wonder if we should change the value of the annotation to `script-runner-image` for consistency if we change the name of the parameter on the task

Zokormazo:
The CI flake8/black issues are interesting, `tox` in local runs without issues:
```
  flake8: OK (0.16=setup[0.03]+cmd[0.13] seconds)
  black: OK (0.22=setup[0.00]+cmd[0.22] seconds)
  test: OK (0.28=setup[0.01]+cmd[0.27] seconds)
  congratulations :) (0.70 seconds)
```

Zokormazo:
Included docstrings and updated the README file, but I found a bug with spdx along the way. `SPDXId` equals to `SPDXRef-image` on builder mode, and must be different. Will fix it asap

Zokormazo:
`SPDXId` issue is fixed. @brunoapimentel , @mmorhun this is ready for final review

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
```diff
@@ -139,16 +166,22 @@ def update_component_in_cyclonedx_sbom(sbom: dict, image: Image) -> dict:
...
         "name": image.name,
         "purl": image.purl(),
         "version": image.tag,
         "hashes": [{"alg": image.digest_algo_cyclonedx, "content": image.digest_hex_val}],
     }
-    sbom["components"].insert(0, image_component)
-    sbom["metadata"]["component"] = image_component
+    if image.is_builder_image:
+        # Add builder image property to image_component
+        image_component["properties"]: [BUILDER_IMAGE_PROPERTY]
```
brunoapimentel, line 178: I caught a last minute bug :sweat:

```suggestion
        image_component["properties"] = [BUILDER_IMAGE_PROPERTY]
```

Otherwise, the property doesn't get rendered in the SBOM.

brunoapimentel: The unit test needs to be fixed as well.

brunoapimentel:
Sorry for the delay in reviewing. This LGTM, we only need to fix the tiny bug I mentioned.

sbom-utility-scripts/scripts/add-image-reference-script/add_image_reference.py
Zokormazo: Interesting, I made the same stupid mistake on the test and that's why the test did not catch it. Thanks, great catch @brunoapimentel !

Fixed

brunoapimentel (approved):

PR #257, opened 2025-03-13
Added missing signatures argument

fmudry:
assure that build won't fail, when tekton task parameter 'IGNORE_MISSING_SIGNATURES' is set to "true" while running skopeo_copy

This is the implementation of the original [PR](https://github.com/konflux-ci/build-tasks-dockerfiles/pull/240), which will enable users to build when source image is missing signatures

This PR works hand in hand with: https://github.com/konflux-ci/build-definitions/pull/2054

### DIFF
--- a/source-container-build/app/source_build.py
+++ b/source-container-build/app/source_build.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python3.11
 
 import argparse
+import subprocess
 import filetype
 import functools
 import hashlib
@@ -56,6 +57,10 @@ class BuildResult(TypedDict):
     base_image_source_included: bool
     image_url: str
     image_digest: str
+
+
+class NoSignatureError(Exception):
+    pass
 
 
 @dataclass
@@ -166,6 +171,11 @@ def parse_cli_args():
         help="Resolve source images for parent images pulled from the registry listed here. "
         "One registry per line.",
     )
+    parser.add_argument(
+        "--ignore-unsigned-image",
+        action="store_true",
+        help="When provided, the build will not fail when source image has missing signatures",
+    )
     return parser.parse_args()
 
 
@@ -194,7 +204,10 @@ def fetch_image_manifest_digest(image: str) -> str:
 
 
 def skopeo_copy(
-    src: str, dest: str, digest_file: str = "", remove_signatures: bool = False
+    src: str,
+    dest: str,
+    digest_file: str = "",
+    remove_signatures: bool = False,
 ) -> None:
     """Execute skopeo-copy
 
@@ -211,7 +224,17 @@ def skopeo_copy(
         flags.append("--remove-signatures")
     cmd = ["skopeo", "copy", *flags, src, dest]
     logger.debug("copy image: %r", cmd)
-    run(cmd, check=True)
+
+    try:
+        run(cmd, stderr=subprocess.PIPE, check=True)
+    except CalledProcessError as e:
+        is_missing_signatures = (
+            "Source image rejected: A signature was required, but no signature exists"
+            in e.stderr.decode()
+        )
+        if is_missing_signatures:
+            raise NoSignatureError(e.stderr.decode())
+        raise
 
 
 # produces an artifact name that includes artifact's architecture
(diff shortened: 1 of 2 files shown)


tkdchen (commented):
As mentioned in the original story, the root cause is some source images are not signed. So, how about using "unsigned image" instead of "missing signature"? To me, the former is much straightforward and points to the root cause directly. For example, the cli arg `ignore_missing_signature` can be renamed to `ignore_unsigned_image`.

source-container-build/app/source_build.py
```diff
@@ -166,6 +171,11 @@ def parse_cli_args():
         help="Resolve source images for parent images pulled from the registry listed here. "
         "One registry per line.",
     )
+    parser.add_argument(
+        "--ignore-missing-signatures",
+        action="store_true",
+        help="When provided, the build will not fail when when source image has mising signatures",
+    )
```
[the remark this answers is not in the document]
fmudry: I missed that, thank you!

source-container-build/app/source_build.py
```diff
@@ -58,6 +59,10 @@ class BuildResult(TypedDict):
     image_digest: str
 
 
+class NoSignatureException(Exception):
+    pass
```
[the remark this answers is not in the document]
fmudry: Thank you, I will make sure not to repeat this mistake.

source-container-build/app/source_build.py
```diff
@@ -507,7 +545,20 @@ def download_parent_image_sources(source_image: str, work_dir: str) -> str:
...
+            remove_signatures=True,
+            ignore_unsigned_image=ignore_unsigned_image,
+        )
+    except NoSignatureError:
+        logger.warning("Source image rejected: A signature was required, but no signature exists")
+        logger.info("Deleting", os.path.join(work_dir, "parent_image_sources"))
+        shutil.rmtree(os.path.join(work_dir, "parent_image_sources"))
+        return ""
+
     return sources_dir
```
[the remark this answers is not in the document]
fmudry: My bad, you are right. I've added unit tests and changed the `download_parent_image_sources` method as well as `skopeo_copy`. Please, let me know if there are more changes to be made.

source-container-build/app/source_build.py
```diff
@@ -507,7 +536,29 @@ def download_parent_image_sources(source_image: str, work_dir: str) -> str:
...
+                "parent_image_sources",
+            ),
+        )
+        shutil.rmtree(os.path.join(work_dir, "parent_image_sources"))
+        if ignore_unsigned_image:
+            logger.info("Ignored missing source image signatures")
+            return ""
+
+        logger.warning("Source image rejected: A signature was required, but no signature exists")
+        raise
```
chmeliik, line 560: Rather than log a hardcoded warning message, IMO we should use the stderr of the failed skopeo process
```suggestion
        raise
```
(tied to the next suggestion)

source-container-build/app/source_build.py
```diff
@@ -211,7 +224,17 @@ def skopeo_copy(
...
+
+    try:
+        run(cmd, stderr=subprocess.PIPE, check=True)
+    except CalledProcessError as e:
+        is_missing_signatures = (
+            "Source image rejected: A signature was required, but no signature exists"
+            in e.stderr.decode()
+        )
+        if is_missing_signatures:
+            raise NoSignatureError
```
chmeliik, line 236:
```suggestion
            raise NoSignatureError(e.stderr.decode())
```
(tied to the previous suggestion)

source-container-build/app/test_source_build.py
```diff
@@ -1292,3 +1292,28 @@ def test_source_image_does_not_exist(self, mock_run: MagicMock):
...
+            "mock cmd",
+            stderr="Source image rejected: A signature was required, but no signature exists".encode(),  # noqa: E501 (ignore line too long)
+        )
+
+    @patch("source_build.run")
+    def test_source_image_is_unsigned_and_not_ingored(self, run: MagicMock):
+        run.side_effect = self.called_proc_err
+
+        with pytest.raises(source_build.NoSignatureError):
+            source_build.download_parent_image_sources("", "", False)
```
chmeliik, line 1311: nitpick: this would be much more obvious at first sight
```suggestion
            source_build.download_parent_image_sources("", "", ignore_unsigned_image=False)
```

chmeliik (commented):
Logic LGTM, just nitpicks

source-container-build/app/source_build.py
fmudry: Thanks for the suggestion!

source-container-build/app/source_build.py
fmudry: Thanks for the suggestion!

source-container-build/app/test_source_build.py
fmudry: Thanks for the suggestion!

PR #269, opened 2025-03-28
[ISV-5787] Remove child digests from externalRefs

mavaras:


mavaras:
@jedinym please take a look. Thanks

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
```diff
@@ -56,30 +56,35 @@ def digest_hex_val(self) -> str:
         return val
 
     def purls(self, index_digest: Optional[str] = None) -> list[str]:
```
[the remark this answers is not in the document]
mavaras: PR updated

BorekZnovustvoritel:
> @BorekZnovustvoritel What is the process on getting this into usage? Do we need to merge and wait for the image to be built, then change the digest in build-definitions?

Yes, I think so. Refer to [this PR](https://github.com/konflux-ci/build-definitions/pull/1513/files#diff-ae600d59a24d6aa921e91738e1cb831631ebebe167d9c2825cf286af9faa9a01R166)

BorekZnovustvoritel (approved):
LGTM, this should work fine.

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
```diff
@@ -55,31 +55,36 @@ def digest_hex_val(self) -> str:
...
+            type="oci",
+            name=self.name,
+            version=index_digest,
+            qualifiers={"arch": self.arch, "repository_url": self.repository},
+        ).to_string()
+
+        # HACK: There's a bug in PackageURL python that incorrectly handles
+        # encoding of ':' characters.  When this PR is merged, the hack should be
+        # removed: https://github.com/package-url/packageurl-python/pull/178
+        return [self._hack_purl_encoding(purl)]
```
chmeliik, line 69: What issues was this causing? Could you separate the change into its own commit with an explanation why it's needed? (If it really is needed. See second comment.)

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
```diff
@@ -55,31 +55,36 @@ def digest_hex_val(self) -> str:
...
 
     def propose_spdx_id(self) -> str:
         purl_hex_digest = hashlib.sha256(self.purls()[0].encode()).hexdigest()
         return f"SPDXRef-image-{self.name}-{purl_hex_digest}"
 
+    def _hack_purl_encoding(purl: str) -> str:
+        """
+        Encode ':' characters in PURL that are not the scheme and type separator.
+        """
```
chmeliik, line 78: Is that correct? According to the [purl spec](https://github.com/package-url/purl-spec/blob/main/PURL-SPECIFICATION.rst#character-encoding):

> the ':' scheme and type separator does not need to and must NOT be encoded. It is unambiguous unencoded everywhere

chmeliik (commented):
@mavaras could you please update the commit message(s) to explain what is being done and why? I have no context for this

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
jedinym: No issues other than that it's not in accordance with the spec. A separate commit is a good idea.

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
jedinym: Yes, the separator must not be encoded. However the rest of the colon symbols _should_ be. See this issue: https://github.com/package-url/packageurl-python/issues/152

jedinym: There's a PR that fixes this issue in package-url python: https://github.com/package-url/packageurl-python/pull/178

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
mavaras: Commits separated. PR updated

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
chmeliik: I went through the issue, the PR and the linked work on the purl-spec side and I'm not convinced. The prevailing opinion seems to be that `:` should never be percent-encoded, period. The purpose of the packageurl-python PR changed from fixing the encoding to fixing the decoding (work correctly even if the name contains a `:`)

chmeliik: The purl-spec PR rabbit hole leads to https://github.com/package-url/purl-spec/pull/439, which clarifies the [percent-encoding rules](https://github.com/package-url/purl-spec/blob/438-clarify-character-encoding/PURL-SPECIFICATION.rst#percent-encoding-rules): 

> When percent-encoding is required, all characters MUST be encoded except for the colon ':'.

mavaras: So what do you suggest? Wait for this to be merged until the package-url PR gets merged?

chmeliik: I would suggest to just drop the second commit.

jedinym: Agreed. Thanks @chmeliik for your insight. I will try and get the [security data guidelines](https://github.com/RedHatProductSecurity/security-data-guidelines/tree/main/sbom/examples/container_image/build) updated.

mavaras: Thanks, PR updated.

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
```diff
@@ -55,26 +55,15 @@ def digest_hex_val(self) -> str:
...
-                qualifiers={"repository_url": self.repository},
-            ).to_string()
-        )
-        return ans
+    def purls(self, index_digest: str) -> list[str]:
+        purl = PackageURL(
+            type="oci",
+            name=self.name,
+            version=index_digest,
+            qualifiers={"arch": self.arch, "repository_url": self.repository},
```
chmeliik, line 63: `self.arch` can be `None` here. PackageURL does the "right thing" and skips falsy values, but to avoid confusion / problems with misusing the API (`qualifiers` are annotated as a `dict[str, str]`), IMO we should only add arch when not None.

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
```diff
@@ -55,26 +55,15 @@ def digest_hex_val(self) -> str:
...
+        ).to_string()
+
+        return [purl]
 
     def propose_spdx_id(self) -> str:
         purl_hex_digest = hashlib.sha256(self.purls()[0].encode()).hexdigest()
```
chmeliik, line 69: We're missing an argument for `self.purls()` here.

I think the script deserves a small refactor to better reflect what it does now. Previously, we would create image objects with different `arch` and `digest` and pass the `image_index_digest` to the purls method to get more purls. We don't really want to do that anymore - IMO all the `Image` objects we create should have the same digest (the `image_index_digest`) and differ only in `arch`

chmeliik: And the `purls()` method should take no parameters, it should just use `self.digest`

sbom-utility-scripts/scripts/index-image-sbom-script/index_image_sbom_script.py
```diff
@@ -55,33 +55,26 @@ def digest_hex_val(self) -> str:
...
+            qualifiers["arch"] = self.arch
+
+        purl = PackageURL(
+            type="oci",
+            name=self.name,
+            version=self.digest,
+            qualifiers=qualifiers,
+        ).to_string()
+
+        return [purl]
```
jedinym, line 70: This function always returns a single value, let's ditch the list.

mavaras: PR updated

jedinym (approved):
lgtm

chmeliik (commented):
Tests failed, PTAL. I think that we're currently using the arch-specific digests instead of the index image digests as intended

mavaras:
> Tests failed, PTAL. I think that we're currently using the arch-specific digests instead of the index image digests as intended
> 
> Btw, you can run tests with `tox` locally

Tests updated. Passing locally

chmeliik:
> Tests updated. Passing locally

Ah, I think it was the script code that needed an update, not the tests. Now, the test data has different digests in the purl for the `SPDXRef-image-index` and in the child image purls, but the digests should be the same

chmeliik (approved):

jedinym:
LGTM

chmeliik:
quay.io is not too stable these days, huh

/retest

jedinym:
another 502
/retest

PR #275, opened 2025-04-01
fix(KONFLUX-7507) Allow content set to be specified in build

arewm:
When we add the content-sets.json file, we also add an additional layer to the image. If any user has expectations about layer content, adding this layer will violate assumtions. We still want to ensure that the content-sets.json file is accurate, but we can allow users to add it on their own and just validate its accuracy.

### DIFF
--- a/icm-injection-scripts/scripts/inject-icm.sh
+++ b/icm-injection-scripts/scripts/inject-icm.sh
@@ -15,6 +15,7 @@
 set -euo pipefail
 
 IMAGE="${1}"
+MOUNT_PATH="${2}"
 SQUASH="${SQUASH:-false}"
 
 icm_filename="content-sets.json"
@@ -69,7 +70,15 @@ done <<< "$(
 echo "Constructed the following:"
 cat content-sets.json
 
-echo "Writing that to $location"
+echo "Checking to see if the content-sets.json file exists in ${location}"
+status=0
+diff -EZbB "${MOUNT_PATH}/${location}" content-sets.json || status=$?
+if [ "$status" -eq 0 ]; then
+  echo "content-sets.json already exists in container and it matches. Skipping injection."
+  exit 0
+fi
+
+echo "No matching content-sets.json found at $location. Adding file into the image."
 buildah copy "$CONTAINER" content-sets.json /usr/share/buildinfo/
 buildah config -a "org.opencontainers.image.base.name=${base_image_name}" -a "org.opencontainers.image.base.digest=${base_image_digest}" "$CONTAINER"
 
--- a/icm-injection-scripts/scripts/test-inject-icm.sh
+++ b/icm-injection-scripts/scripts/test-inject-icm.sh
@@ -29,7 +29,7 @@ cd "$WORKDIR"
 
 banner "Running inject-icm.sh with a $TEST_SBOM_FORMAT SBOM"
 cp "$SCRIPTDIR/test-data/sbom-cachi2-$TEST_SBOM_FORMAT.json" ./sbom-cachi2.json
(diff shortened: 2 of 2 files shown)


icm-injection-scripts/scripts/inject-icm.sh
```diff
@@ -69,7 +70,15 @@ done <<< "$(
 echo "Constructed the following:"
 cat content-sets.json
 
-echo "Writing that to $location"
+echo "Checking to see if the content-sets.json file exists in ${location}"
+status=0
+diff -EZbB ${MOUNT_PATH}/${location} content-sets.json || status=$?
```
mmorhun, line 75: I think we need to use `"` for `${MOUNT_PATH}/${location}` to be safe.

mmorhun (approved):

icm-injection-scripts/scripts/inject-icm.sh
```diff
@@ -15,6 +15,7 @@
 set -euo pipefail
 
 IMAGE="${1}"
+MOUNT_PATH="${2}"
```
chmeliik, line 18: Mount path is a path to the exported filesystem of the image? We want to stop exporting that (https://issues.redhat.com/browse/STONEBLD-3020, which may need to be fixed as part of https://github.com/konflux-ci/build-definitions/pull/2074/files)

arewm: Yes, I was planning to push a change to the buildah tasks to modify the execution to
```
/scripts/inject-icm.sh "$IMAGE" "$(cat /shared/container_path)"
```

Do you have another thought for how we can do this? When we stop mounting the container, we could introduce a change to this script to mount (and clean up with unmounting) the container instead.

chmeliik: Alternatively, we would need a tool that knows how to extract a single file from an OCI container. Maybe `oc extract` would work?

chmeliik: I noticed script already uses `buildah`, so yeah, mounting temporarily in this script should work. Or doing something like `buildah run $container -- cat /root/buildinfo/...`

arewm: The options that I see right now are:

* We leave this script as it is, consistently changing the behavior when we address the issue with syft
* If we pass `$container` to this script, we can mount it here and that will return a consistent mount path for all mount operations.
* If we just pass `$IMAGE` to the script, we can re-run `buildah from --pull-never "$IMAGE"` within this script but that will create a new working container and mounting will duplicate any file system mounting.
* We could pass the `$container` in as an alternative. I don't think we should run the container as we cannot guarantee that `cat` is present nor can we guarantee that we should trust that binary.

What are your thoughts about the approach we should take?

chmeliik: > We leave this script as it is, consistently changing the behavior when we address the issue with syft

Adressing the issue with syft will not involve mounting the container FS or otherwise extracting any individual files, so consistency wouldn't be a factor here.

> we cannot guarantee that cat is present nor can we guarantee that we should trust that binary.

Good point :+1: 

> If we just pass `$IMAGE` to the script, we can re-run `buildah from --pull-never "$IMAGE"` within this script but that will create a new working container and mounting will duplicate any file system mounting.

Yeah, we can either pass `$container` (if it still exists after the overlayfs changes and works across task steps) or `$IMAGE`

Do you think this PR is higher priority than https://github.com/konflux-ci/build-definitions/pull/2074? It feels like the performance improvements from overlayfs would have much higher impact and I would wait for that one to get in first than vice versa

PR #305, opened 2025-05-21
add retries for skopeo commands

rcerven:
STONEBLD-3488


mmorhun (approved):

source-container-build/app/source_build.py
```diff
@@ -46,6 +47,8 @@
 )
 
 MAX_RETRIES: Final = 5
+SUBPROCESS_BACKOFF_FACTOR: Final = 2
+SUBPROCESS_MAX_RETRIES: Final = 10
```
chmeliik, line 51: We are now calling `skopeo ... --retry {MAX_RETRIES}` and then retrying all of that 10 times, which seems like a big overkill

IMO we should:
* Not retry subprocesses 10 times. With exponential backoff with base=2 and factor=2, that would mean up to 34 minutes of waiting
* Either
  * stop using the `--retry` option
  * or rely on the `--retry` option and not introduce more retries in the code

chmeliik: Discussed offline: we still want `--retry` as a fast retry mechanism while adding `backoff.on_exception` for cases where that's not enough or where skopeo fails for a reason that doesn't cause a retry

PR #316, opened 2025-06-05
Rebase source container image on fedora

ralphbean:
The point is to get the new skopeo 1.19.

It includes this fix in containers/common: https://github.com/containers/common/pull/2378

Once skopeo 1.19 is released for ubi9, we can switch back to ubi9.

### DIFF
--- a/source-container-build/Dockerfile
+++ b/source-container-build/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/ubi:9.6-1747219013
+FROM quay.io/fedora/fedora:latest@sha256:038a2cc23755fe0fd387587c0cb075c044d18bca103bdff56b46e6b3c564f9f7
 
 ARG BSI_VERSION=0.2.0
 ARG bsi_source=https://github.com/containers/BuildSourceImage/archive/refs/tags/v${BSI_VERSION}.tar.gz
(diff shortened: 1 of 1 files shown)


ralphbean:
Note, skopeo 1.19 isn't yet in the stable updates repo for Fedora, but it should be there by tomorrow: https://bodhi.fedoraproject.org/updates/FEDORA-2025-01bc496b18

rcerven:
> @rcerven PTAL, now with the skopeo retries fix would we still need extra retries in the code? #305

I'd leave it there, it doesn't hurt

also I'd say that skopeo retries will have very low wait time

ralphbean:
Real question: do we want to build our own skopeo binary using the submodule pattern like we do for yq rather than rebase on fedora here?

chmeliik:
> Real question: do we want to build our own skopeo binary using the submodule pattern like we do for yq rather than rebase on fedora here?

+1, but it may need some care. If nothing else, we should probably keep installing the `containers-common` RPM (a dependency of skopeo) because it comes with various container-related config files. Without those, we may change the behavior in unexpected ways (e.g. disable signature verification?)

And rather than submodules, I would go with a simple go.mod listing `skopeo` as the dependency. Same approach as was used for syft and yq in https://github.com/redhat-appstudio/tssc-dev-multi-ci/pull/114/files

ralphbean:
At this point, I'm thinking that rebasing on top of Fedora is not the right way to go here. I'm going to drop this. (If you disagree, feel free to re-open it and we can restart conversation.)

PR #336, opened 2025-06-30
Check binary image existence when ready to push

tkdchen:
KONFLUX-8161

When ready to push the built source image, make sure the binary image is still valid. This behavior addresses a scenarior using floating tag in Component image repositories. Here is an example of output-image passed to build pipeline:

  - name: output-image
    value: quay.io/org/registry-operator:{{target_branch}}

this results in potential race condition when two PipelineRuns run in parallel and the latest source container image can be overwritten by the older one built by the other PipelineRun.

To solve this issue, this solution introduces the following major changes:

* --output-binary-image is renamed to --binary-image-ref which must accept a full image reference with both tag and digest.
* A new class ImageRef helping validating and constructing kinds of image reference on-demand.
* Check input binary image just before pushing the source image
* Build result has a new status named "drop". Then, when the input binary image does not exist anymore in the registry, stop the process immediately without push and build result has status "drop" and a meaningful message.

### DIFF
--- a/source-container-build/app/source_build.py
+++ b/source-container-build/app/source_build.py
@@ -54,7 +54,7 @@ StrPath = str | os.PathLike
 
 
 class BuildResult(TypedDict):
-    status: Literal["failure", "success"]
+    status: Literal["failure", "success", "drop"]
     message: NotRequired[str]
     dependencies_included: bool
     base_image_source_included: bool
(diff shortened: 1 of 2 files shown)


tkdchen:
e2e tests is run by pairing with https://github.com/konflux-ci/build-definitions/pull/2489. Please check e2e tests status in that PR.

chmeliik (approved):

tkdchen:
Retrigger PipelineRuns so that the image is built again and accessible by the CI in https://github.com/konflux-ci/build-definitions/pull/2489

PR #413, opened 2026-04-16
Use git-core with ubi9.7

MartinBasti:
New UBI image brought some perl dependency issues, however we don't need perl deps at all, we need only git-core


MartinBasti:
fixes https://github.com/konflux-ci/build-tasks-dockerfiles/pull/344

mmorhun (approved):

PR #417, opened 2026-04-23
build: replace appstudio-utils with task-runner

fmudry:
`appstudio-utils` image will be decommissioned soon and the recommended replacement is the `task-runner` image.

The change should be tested by the CI.

### DIFF
--- a/integration-tests/pipelines/e2e-tests-pipeline.yaml
+++ b/integration-tests/pipelines/e2e-tests-pipeline.yaml
@@ -48,12 +48,19 @@ spec:
                 memory: 4Gi
               limits:
                 memory: 6Gi
-            image: quay.io/konflux-ci/appstudio-utils:4b2d7c21c248df18a523f0e210f345eb463d61c1@sha256:a62a98368ac9d8f13026bc928510e0757d7242048797440916a1564eab5256c7
+            image: quay.io/konflux-ci/task-runner:1.6.0@sha256:1abfe4e50d4e961d0fd9790202565f93ee650fe8dfc50932c94989acba10485f
+            securityContext:
+              runAsUser: 0
+              runAsNonRoot: false
             env:
             - name: SNAPSHOT
               value: $(params.SNAPSHOT)
             script: |
-              set -e
+              set -euo pipefail
+
+              # Go is required by the e2e-tests repo (mage build tool), but (as of now) is not included in task-runner
+              microdnf -y install golang
+
               echo "SNAPSHOT: ${SNAPSHOT}"
               export SOURCE_BUILD_IMAGE=$(jq -r '.components[] | select(.name == "source-container-build") | .containerImage' <<< "$SNAPSHOT")
               git clone --branch main https://github.com/konflux-ci/e2e-tests.git


fmudry:
it seems that the task-runner doesn't have mage:
```
  Cloning into 'e2e-tests'...
  Setup pipeline bundle replacing source build image
  ./mage -v SetupBundleForBuildTasksDockerfilesRepo
  ./mage: line 8: go: command not found
  make: *** [Makefile:91: setup-only-source-build] Error 127
  Failed to read pipeline bundle, exiting..
```

fmudry:
the `prepare-e2e-tests` step passes now :tada:

integration-tests/pipelines/e2e-tests-pipeline.yaml
```diff
@@ -56,7 +56,16 @@ spec:
...
-              set -e
+              set -euo pipefail
+
+              # Go is required by the e2e-tests repo (mage build tool), but (as of now) is not included in task-runner
+              echo "Installing Go"
+              export GOVERSION=1.26.2
+              curl -sLO "https://go.dev/dl/go${GOVERSION}.linux-amd64.tar.gz"
+              tar -C /usr/local -xzf "go${GOVERSION}.linux-amd64.tar.gz"
+              rm "go${GOVERSION}.linux-amd64.tar.gz"
+              export PATH="/usr/local/go/bin:${PATH}"
```
chmeliik: Would `microdnf -y install golang` work, or do we need specifically 1.26? We should avoid downloading unverified binaries if possible.

integration-tests/pipelines/e2e-tests-pipeline.yaml
```diff
@@ -48,7 +48,10 @@ spec:
                 memory: 4Gi
               limits:
                 memory: 6Gi
-            image: quay.io/konflux-ci/appstudio-utils:4b2d7c21c248df18a523f0e210f345eb463d61c1@sha256:a62a98368ac9d8f13026bc928510e0757d7242048797440916a1564eab5256c7
+            image: quay.io/konflux-ci/task-runner:1.6.0@sha256:1abfe4e50d4e961d0fd9790202565f93ee650fe8dfc50932c94989acba10485f
+            securityContext:
+              runAsUser: 0
+              runAsNonRoot: false
```
chmeliik, line 54: nitpick: the reason why this is necessary doesn't yet exist in this commit. They should probably be squashed, since the first one on its own wouldn't pass CI

integration-tests/pipelines/e2e-tests-pipeline.yaml
chmeliik: Actually, if we do `export GOTOOLCHAIN=auto GOSUMDB=sum.golang.org`, then older versions should automatically fetch the newer version, in a slightly safer way. It at least checks the download against the checksum server, though we're still trusting the checksum server

fmudry: Let me verify :)

integration-tests/pipelines/e2e-tests-pipeline.yaml
fmudry: Good point, it makes more sense :+1: squashed

integration-tests/pipelines/e2e-tests-pipeline.yaml
fmudry: Yeah, seems to be working. I think using `microdnf` is much better, thanks!

chmeliik (approved):

